PR #2, opened 2020-06-23
Add empty html files for hard coded front end.

ganewman:

Add empty html files for the directory and candidate pages.

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [  ] Tests passing
- [  ] Coding style (indentation, etc)


project/src/main/webapp/directory.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8">
+    <meta name="Voter Central Candidate Directory Page"
```
kellyktan: Would it make more sense to call this a "election directory page" instead of "candidate", since it's not really surfacing individual candidates at this point?

project/src/main/webapp/candidate.html
```diff
@@ -0,0 +1,15 @@
...
+          content="The sample candidate page of Voter Central, a website
+                   for centralizing voting information">
+    <link rel="stylesheet" href="style.css">
+    <script src="config.js"></script>
+    <script src="script.js"></script>
+    <title>Voter Central Candidate Page</title>
+  </head>
+  <body>
+  </body>
+</html>
```
KathyFeiyang: nit: Add a newline at the end of the file.

project/src/main/webapp/directory.html
```diff
@@ -0,0 +1,15 @@
...
+          content="The directory page of Voter Central, a website
+                   for centralizing voting information">
+    <link rel="stylesheet" href="style.css">
+    <script src="config.js"></script>
+    <script src="script.js"></script>
+    <title>Voter Central Directory Page</title>
+  </head>
+  <body>
+  </body>
+</html>
```
KathyFeiyang: nit: Add a newline at the end of the file.

KathyFeiyang (changes requested):

project/src/main/webapp/directory.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8">
+    <meta name="Voter Central Candidate Directory Page"
+          content="The directory page of Voter Central, a website
```
suparnabarua, line 6: You shouldn't have to describe what Voter Central is in every file. Just one file, perhaps the landing page, should be enough.

kellyktan: I think this isn't actually shown on a page, but is more of an accessibility label?

suparnabarua: Didn't see that before. But looks like the content tag should be replaced with slot since the former is deprecated
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/content

kellyktan: jk I think the meta tag is more of a search engine thing that indexers look at?

ganewman: > Didn't see that before. But looks like the content tag should be replaced with slot since the former is deprecated
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/content

I think that might be for the `<content>` tag, rather than the content attribute of the `<meta>` tag.

KathyFeiyang: I believe it's used to specify metadata about the page: https://www.w3schools.com/tags/tag_meta.asp. The content attribute for the meta element appears to be different from the content element.

KathyFeiyang (approved):
LGTM

tjyoung99 (approved):

kellyktan (approved):

PR #4, opened 2020-06-25
Add basic styling and common elements.

ganewman:

- Add some basic divs to the candidate page, and use CSS to define their size. 
- Add the JS code which will insert the header and sidebar onto every page of the website. Add HTML to support this, and some basic CSS to position the sidebar and make the links align.

### Is this fixing an issue?

none

An additional JS file, `common.js`, has been added to store code specifically related to the header and sidebar. This will become necessary as the sidebar will eventually contain dynamic dropdown menus.

- [  ] Test written/updating
- [X] Tests passing
- [X] Coding style (indentation, etc)

### Any additional comments?

{Fill in here or remove (leave question)}


project/src/main/webapp/candidate.html
```diff
@@ -8,9 +8,20 @@
...
     <script src="script.js"></script>
+    <script src="common.js"></script>
     <title>Voter Central Candidate Page</title>
   </head>
   <body>
-    <h2>Voter Central Candidate Page</h3>
+    <div id="header-container"></div>
+    <div id="navbar-container"></div>
+    <h2>Voter Central Candidate Page</h2>
+    <div id="content">
```
KathyFeiyang: `<div>` doesn't have enough semantic meaning. Maybe change it to `<main>`.

project/src/main/webapp/candidate.html
```diff
@@ -8,9 +8,20 @@
     <link rel="stylesheet" href="style.css">
     <script src="config.js"></script>
     <script src="script.js"></script>
+    <script src="common.js"></script>
```
KathyFeiyang: The name "common" might be changed to something more descriptive of the functionality, such as "navigation"

project/src/main/webapp/common.js
```diff
@@ -0,0 +1,38 @@
...
+  addHeader();
+}
+
+function addNavbar() {
+  let navbar = `<a href="/index.html">Home</a>
+      <br>
+      <a href="/">Resources</a>
+      <br>
+      <a href="/">Polling Locations</a>
+      <br>`;
```
KathyFeiyang: The code may be more maintainable if you use JavaScript to create elements, instead of putting everything in a String. Especially if the navbar HTML structure will change/expand in the future.
If the HTML structure will unlikely change and the hrefs are the only thing that will be filled out, then using a String is more convenient.

project/src/main/webapp/common.js
```diff
@@ -0,0 +1,38 @@
...
+  let navbarDiv = document.getElementById('navbar-container');
```
KathyFeiyang: nit: can use `const navbarDiv`

project/src/main/webapp/common.js
```diff
@@ -0,0 +1,38 @@
...
+  navbarDiv.innerHTML = navbar;
+}
+
+function addHeader() {
+  let header = `<h1>Voter Central</h1>`;
+  let headerDiv = document.getElementById('header-container');
```
KathyFeiyang: nit: use const whenever possible. headerDiv remains the same object.

project/src/main/webapp/common.js
```diff
@@ -0,0 +1,38 @@
...
+      <a href="/">Resources</a>
+      <br>
+      <a href="/">Polling Locations</a>
+      <br>`;
+  let navbarDiv = document.getElementById('navbar-container');
+  navbarDiv.innerHTML = navbar;
+}
+
+function addHeader() {
+  let header = `<h1>Voter Central</h1>`;
```
KathyFeiyang: Same comments as navbar above.
Additionally, if header will be just one HTML element, we can put it directly in the HTML.

project/src/main/webapp/candidate.html
KathyFeiyang: Just a reminder: the corresponding CSS styling rule would need to change as well

KathyFeiyang (changes requested):

suparnabarua (approved):
LGTM pending resolution of existing comments

project/src/main/webapp/common.js
ganewman: Using the JS to create elements this long tends to make a bit of an unreadable mess. I think my plan is to create elements for the dynamic parts, but leave the hardcoded links as a string.

KathyFeiyang: Ok! If this section will remain the fixed structure, I agree it's easier to have a hard-coded string. You can fill in the links later with string formatting.

project/src/main/webapp/common.js
ganewman: I was planning on possibly adding some links to it later, so I separated it out as well just in case.

project/src/main/webapp/style.css
```diff
@@ -3,4 +3,41 @@ body {
...
+  text-align: center;
+  width: 15%;
+}
+
+#navbar-container a {
+  padding: 10px 20px;
+  vertical-align: middle;
+}
+
+
```
kellyktan: nit: extra line?

project/src/main/webapp/common.js
KathyFeiyang: Got it, sounds great!

kellyktan (approved):
LGTM pending open comments

project/src/main/webapp/style.css
ganewman: Oops.

KathyFeiyang (approved):
LGTM!

tjyoung99 (approved):

PR #8, opened 2020-06-30
Implement frontend, backend and database communication for location-based search results (directory page) and the news articles component (candidate page). Query for and serve dynamic content in designed HTML structure.

KathyFeiyang:

_**Implement frontend, backend and database communication for location-based search results (directory page) and the news articles component (candidate page). Query for and serve dynamic content in designed HTML structure.**_

**1. Implement backend Java servlets for querying the database for:**
  (1) Location-based search results, which are brief election and candidate information (directory page).
  (2) News articles component (candidate page).
- Determine the Java data structures for holding the dynamic content.
- For content 1, the `DataServlet` organizes the raw database data into a data structure that better fits the designed HTML structure: One election corresponds to a list of positions, whereas a position corresponds to a table of candidates and their associated information.
- The `DataServlet` will need to call the Civic Information API to resolve an address to one or more election names. Currently, it obtains user input of address from the frontend, but queries for a hard-coded election name.
- The `CandidateServlet` also contains skeleton code for the official information and social media components on the candidate page.
- The code reflects a change in the database schema, where elections are now keyed by election names. The notion of a political unit of location is removed from the database. It is now only used as the query start word for compiling information.
 

**2. Implement frontend JavaScript code for requesting 2 categories of dynamic content (as listed above), and constructing said content with the designed HTML structure.**
- Handle the passing of parameters between HTML pages:
    1. Pass user input of address from the homepage to the directory page, to perform location-based search.
    2. Pass a candidate ID from the directory page's clickable candidate name to the candidate page, to retrieve candidate-specific information.
- Include skeleton code for inserting the official information and social media components into the HTML structure.
- Fix issues with and adjust the HTML template structure for content 1.
- Adjust the homepage to take in a textual input of address, instead of a zip code.

**3. Live server:** [go/voter-central](http://go/voter-central)

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [ x ] Tests passing


project/src/main/webapp/script.js
```diff
@@ -20,17 +20,166 @@ window.onload = function() {
...
+    electionElement.appendChild(positionsElement);
+    electionsContainer.appendChild(electionElement);
+  }
+}
+
+/**
+ * Adds candidate information to the candidate page.
+ */
+async function addCandidateInformation() {
+  console.log(location.search);
```
suparnabarua: Remove this

project/src/main/webapp/script.js
```diff
@@ -20,17 +20,166 @@ window.onload = function() {
...
+  // Extract candidate ID.
```
suparnabarua: Remove this comment

project/src/main/webapp/script.js
```diff
@@ -20,17 +20,166 @@ window.onload = function() {
...
+ *         </li>
+ *       </ul>
+ *     </div>
+ *     <hr>
+ *   </div>
  */
-function addOfficialElectionCandidateInformation() {
-  // Capture user input of location.
+async function addBriefElectionCandidateInformation() {
+  // Extract user input of address.
```
suparnabarua: Remove this comment.

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -28,25 +41,75 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // Extracts user input of some location.
+    // Extract user input of an address.
```
suparnabarua: Remove this comment. It's obvious that the code is extracting the address thanks to the good variable name

project/src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,4 +14,56 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class Election {
+  private String name;
+  private Date date;
+  private List<Position> candidatePositions;
+
+  public Election(String inputName, Date inputDate, List<String> inputCandidateIds,
```
suparnabarua: You don't have to name the arguments with 'input' prefixed to them. Doing this.name = name ensures that the class variables get initialized accurately

project/src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,4 +14,56 @@
...
+    this.name = inputName;
+    this.date = inputDate;
+    // Reformat database data to correlate one {@code Position} object with a list of candidates
+    // and their information.
+    Set<String> distinctPositions = new HashSet<>(inputCandidatePositions);
+    this.candidatePositions = new ArrayList<>(distinctPositions.size());
+    for (String position : distinctPositions) {
+      int startIndex = inputCandidatePositions.indexOf(position);
+      int endIndex = inputCandidatePositions.lastIndexOf(position);
+      Position candidatePosition = new Position(position,
```
suparnabarua: TBH I'm not a fan of this design where you're assuming  the indices correspond to the same candidate across the different lists. Have you thought about having a map instead? If not, feel free to reply with your logic for going with this approach :)

suparnabarua (changes requested):

project/src/main/java/com/google/sps/data/Election.java
KathyFeiyang: Yes, I agree that this is not ideal. I thought about having a map. The reason why I used this approach is because manipulating the database to keep an object (election) of object (positions) of object (candidates running for the positions) is complex. I researched into this, and this is doable via Datastore ancestor paths. Alternatively, I can also achieve this by creating an intermediary table for position VS candidates. However, the trade-off would be complexities with (1) correctly formatting everything into the database and (2) querying the database.
Because of these reasons, I decided to go with the easiest approach. But I'm totally open to changing it

kellyktan: Can candidates potentially run for different positions or be part of different elections? Will the candidate information be duplicated in this case?

I feel like having a table of candidates, and positions only having references to the candidate ID (such that you load whatever candidate information separately) might make sense?  Or maybe candidate ID + election/position specific info, and then everything else goes into the common candidate store?

KathyFeiyang: Yes, candidates can run for different positions or be part of different elections. But candidate information won't be duplicated. Currently there is a table of candidates, keyed by ID. There is a table of elections, keyed by election names. Each election entry has a list of positions and a list of candidate IDs. The i-th candidate is running for the i-th position.

KathyFeiyang: Should I create a separate table of position-candidate ID pairs (or say position objects)? Then each election would only contain a list of position IDs. Each position entry would contain a candidate ID, and whether the candidate is an incumbent.

kellyktan: Oh does the database not easily store object lists?

kellyktan: I think the thing that most feels weird is the 4 lists of strings with candidate information instead of 1 list of "candidate" objects.

KathyFeiyang: Yeah the database doesn't easily store list of objects. Lists of IDs are easy though.

Oh I see what you mean by duplication now. Apologies for the confusion, the 4 lists of strings are a brief set of candidate data that are organized in a format that will be easily parsed by the frontend. It is possible that one candidate can appear multiple times. The "id" and "name" property may be duplicated if a candidate runs for multiple positions/elections, while the incumbency is specific to the position. In the database, there is only a reference to the candidate ID, so information is not duplicated in the database.

KathyFeiyang: Would you suggest I instead assemble data in the frontend? Then the frontend would search for candidate information based on ID and then organize the data.

kellyktan: Ah good to know that the candidate info isn't duplicated in the database.  This class is mostly for packaging the information for passing to the front end, correct?  In that case would it be valid to have a structure like:
```
Election(inputName, inputDate, List<Position> inputPositions)
Position(inputName, List<Candidate> candidates)
Candidate(id, name, partyAffiliation, currentPositionId)
```
And actually construct the Candidate and Position objects in DataServlet.java to pass into the Election object?

`Candidate` could also be called something like `DirectoryCandidate` so you know it's only the info for the directory page...or you could just reuse the Candidate object and only populate certain fields if enough overlaps idk

KathyFeiyang: Yes, this class is for packaging information for the front end.

Got it, I'll revise the structure. I'll rename `Candidate` to `DirectoryCandidate` for now.

Would you recommend modifying the database schema as well and create a Position type?

kellyktan: No need to modify the database schema as long as the data is being stored and can be transformed to this form relatively efficiently.

KathyFeiyang: Got it, thanks! I'll start making the changes.

project/src/main/java/com/google/sps/data/DirectoryCandidate.java
```diff
@@ -0,0 +1,29 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+public class DirectoryCandidate {
```
suparnabarua, line 20: Include a description for this class

project/src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,4 +14,20 @@
...
 package com.google.sps.data;
 
-public class Election {}
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class Election {
```
suparnabarua, line 27: Include a short description for this class

project/src/main/java/com/google/sps/data/NewsArticle.java
```diff
@@ -14,8 +14,25 @@
 
 package com.google.sps.data;
 
+import java.util.Date;
+
 public class NewsArticle {
+  private String title;
```
suparnabarua, line 23: Include a short description for this class

project/src/main/java/com/google/sps/data/Position.java
```diff
@@ -14,4 +14,14 @@
 
 package com.google.sps.data;
 
-public class Location {}
+import java.util.List;
+
+public class Position {
```
suparnabarua, line 23: Include a short description for this class

project/src/main/java/com/google/sps/servlets/CandidateServlet.java
```diff
@@ -14,41 +14,87 @@
...
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-@WebServlet("/candidateServlet")
+@WebServlet("/candidate")
```
suparnabarua, line 41: Include a description for this class

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -28,25 +45,84 @@
...
+  private List<Election> findBriefElectionCandidateInformation(List<String> electionNames) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    // Query for (brief version) official election/candidate information, with election names.
+    List<Entity> elections = new ArrayList<>(electionNames.size());
+    for (String electionName : electionNames) {
+      Query electionQuery = new Query("Election")
+          .setFilter(new FilterPredicate("__key__", FilterOperator.EQUAL,
+                    KeyFactory.createKey("Election", electionName)));
+      PreparedQuery electionQueryResult = datastore.prepare(electionQuery);
+      elections.add(electionQueryResult.asSingleEntity());
```
suparnabarua: Instead of having 2 for loops where you first get the election entities and then populate electionsData, you could call a helper function that does what your second for loop does and call it something like extractCandidateInformation. That way you also just have one list electionsData

suparnabarua (changes requested):
Mostly documentation feedback

project/src/main/webapp/script.js
```diff
@@ -20,17 +20,161 @@ window.onload = function() {
...
+    const positionsElement = document.createElement('div');
+    positionsElement.id = electionElement.id + `-positions`;
+    const positionsList = document.createElement('ul');
+    for (let positionIndex = 0; positionIndex < positions.length;
+        positionIndex++) {
+      const position = positions[positionIndex];
+      const positionItem = document.createElement('li');
+      positionItem.innerHTML =
+          `<h4>Position ${positionIndex + 1}: ${position.name}</h4>
+           <p>Table of Candidates</p>`;
```
ganewman: Consider making a helper function that will build a table of candidates, or one that will build a single row for a candidate, so that this loop is a bit easier to read!

project/src/main/webapp/script.js
```diff
@@ -20,17 +20,161 @@ window.onload = function() {
...
+ */
+function addNewsArticles(newsArticles) {
+  const newsArticlesContainer = document.getElementById('news-articles-container');
+  newsArticlesContainer.innerHTML = '';
+  for (let i = 0; i < newsArticles.length; i++) {
+    const newsArticle = newsArticles[i];
+    const articleElement = document.createElement('article');
+    articleElement.id = `news-article-${i + 1}`;
+    articleElement.innerHTML =
+        `<h3>${newsArticle.title}</h3>
```
ganewman, line 204: Maybe you can create a more generic function that replicates this behavior?

ganewman (approved):
LGTM pending other comments and possible changes to code structure!

project/src/main/webapp/script.js
KathyFeiyang: Thanks for making this suggestion. I have broken the function into 5 hierarchical helper functions:

addBriefElectionCandidateInformation
--> construct one election div
-----> construct a list of positions
--------> construct one position list item
-----------> construct a table of candidates
--------------> construct one row of candidate information

project/src/main/webapp/script.js
KathyFeiyang: Thanks for the suggestion! At the moment, I believe the construction news articles/elections/positions/candidates HTML is too irregular. A generic function would require a large number of parameters that might make the function not so generic.
However, I believe this is an important point, and I have made a note of this. I will circle back to see if extracting a generic behavior is suitable, after completing the migration to Angular JS.

project/src/main/java/com/google/sps/servlets/DataServlet.java
KathyFeiyang: Thanks for pointing this out! I've made the changes to `DataServlet`

project/src/main/java/com/google/sps/data/DirectoryCandidate.java
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/** 
+ * Represents one candidate's information on the directory page.
```
suparnabarua: nit. "a candidate's information"

project/src/main/java/com/google/sps/data/Election.java
```diff
@@ -14,4 +14,24 @@
...
 
-public class Election {}
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/** 
+ * Represents one election's information on the directory page. One election contains one or more
```
suparnabarua: nit. "an election's information on the directory page. An election contains..."

project/src/main/java/com/google/sps/data/NewsArticle.java
```diff
@@ -14,8 +14,28 @@
 
 package com.google.sps.data;
 
+import java.util.Date;
+
+/** 
+ * Represents one news article, including its metadata and content.
```
suparnabarua: nit. "a news article"

project/src/main/java/com/google/sps/data/Position.java
```diff
@@ -14,4 +14,18 @@
 
 package com.google.sps.data;
 
-public class Location {}
+import java.util.List;
+
+/** 
+ * Represents one position for which candidates run in an election. One position corresponds
```
suparnabarua: nit. "Represents a position"

project/src/main/webapp/directory.html
```diff
@@ -37,19 +37,19 @@ <h4>Position 2...</h4>
...
           <ul>
             <li>
               <h4>U.S. Representative</h4>
               <p>Table of Candidates</p>
               <table class="candidates-table">
                 <tr>
-                  <th><a href="URL">Candidate Name</a></th>
+                  <th>Candidate Name</th>
                   <th>Party Affiliation</th>
                   <th>Incumbent?</th>
```
suparnabarua: Remove the '?'

suparnabarua (approved):
LGTM pending minor comments

tjyoung99 (approved):

project/src/main/java/com/google/sps/data/DirectoryCandidate.java
```diff
@@ -0,0 +1,32 @@
...
+package com.google.sps.data;
+
+/** 
+ * Represents a candidate's information on the directory page.
+ */
+public class DirectoryCandidate {
+  private String id;
+  private String name;
+  private String partyAffiliation;
+  private String incumbency;
```
kellyktan: nit: Would this make more sense as a bool?

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -62,47 +70,72 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  private List<DirectoryCandidate> extractCandidateInformation(List<String> candidateIds,
+      List<String> candidateIncumbency) {
+    List<DirectoryCandidate> candidates = new ArrayList<>(candidateIds.size());
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    for (int i = 0; i < candidateIds.size(); i++) {
+      String id = candidateIds.get(i);
+      String incumbency = candidateIncumbency.get(i);
+      Query candidateQuery = new Query("Candidate")
+          .setFilter(new FilterPredicate("__key__", FilterOperator.EQUAL,
+                      KeyFactory.createKey("Candidate", Long.parseLong(id))));
```
kellyktan: Is there a way to do a batch query? i.e. pass in the whole list of candidate IDs? Not sure if this is actually costing anything in the way of GCP credits (if not, feel free to ignore).

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -62,47 +70,72 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+   * Formats position and its associated candidates' information as {@code Position}. Correlates
+   * one {@code Position} object with one or more {@code DierctoryCandidate}.
+   */
+  private List<Position> extractPositionInformation(List<String> candidatePositions,
+      List<DirectoryCandidate> candidates) {
+    Set<String> distinctPositions = new HashSet<>(candidatePositions);
+    List<Position> positions = new ArrayList<>(distinctPositions.size());
+    for (String positionName : distinctPositions) {
+      int startIndex = candidatePositions.indexOf(positionName);
+      int endIndex = candidatePositions.lastIndexOf(positionName);
```
kellyktan, line 109: Are you using these anywhere?

kellyktan (changes requested):

project/src/main/java/com/google/sps/data/DirectoryCandidate.java
KathyFeiyang: Yes, I'll make the changes. This decision was made when I mistakenly thought datastore arrays don't support boolean values

project/src/main/java/com/google/sps/servlets/DataServlet.java
KathyFeiyang: It is, but it is a bit complex to execute. I'll look into this

project/src/main/java/com/google/sps/servlets/DataServlet.java
KathyFeiyang: Ah thanks for the reminder! These should have been used to find the candidates corresponding to a particular position. I'll fix this

project/src/main/java/com/google/sps/data/DirectoryCandidate.java
```diff
@@ -21,9 +21,9 @@
   private String id;
   private String name;
   private String partyAffiliation;
-  private String incumbency;
+  private boolean incumbency;
```
kellyktan: nit: change field name to isIncumbent

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -131,7 +131,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     for (String positionName : distinctPositions) {
       int startIndex = candidatePositions.indexOf(positionName);
       int endIndex = candidatePositions.lastIndexOf(positionName);
```
kellyktan: This assumes that all candidates with the same position will be consecutive, right? Maybe state that assumption somewhere.

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -85,7 +85,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       Entity election = electionQueryResult.asSingleEntity();
       List<DirectoryCandidate> candidates =
           extractCandidateInformation((List<String>) election.getProperty("candidateIds"),
-                                      (List<String>) election.getProperty("candidateIncumbency"));
+                                      (List<Boolean>) election.getProperty("candidateIncumbency"));
       List<Position> positions =
           extractPositionInformation((List<String>) election.getProperty("candidatePositions"),
                                      candidates);
```
kellyktan: Instead of doing this in 2 steps, why not just have one function that (internally) constructs a map<position name, list of candidates>?

Then `extractCandidateInformation` could instead just deal with a single candidate and you call that for each loop.

kellyktan (changes requested):

project/src/main/java/com/google/sps/servlets/DataServlet.java
KathyFeiyang: Datastore also supports an "IN" operator for a list of candidate IDs, in addition to "EQUAL". However, each individual ID will count as a query. "IN" also places a restriction that the size of the list must be < 30. This means that I would need to create sublists of size 30. Therefore, this approach probably won't save queries or simplify the code.

KathyFeiyang: There is also a [bulk deletion pipeline](https://cloud.google.com/dataflow/docs/guides/templates/provided-utilities#datastore-bulk-delete). However, this doesn't fit our particular use case

project/src/main/java/com/google/sps/servlets/DataServlet.java
KathyFeiyang: Will do!

project/src/main/java/com/google/sps/servlets/DataServlet.java
KathyFeiyang: Got it! Just to confirm, by "internally" are you referring to a helper function in DataServlet? Would removing the Position class be good? The mapping would represent all information represented in Position.

kellyktan: You can probably put it all in `extractPositionInformation`. Or sure, removing position could work as well~

KathyFeiyang: Got it, thanks! I'll refactor the code for `extractPositionInformation` and `extractCandidateInformation`
I thought you meant a literal map<position name, list of candidates>. If not, I can just keep the Position class, since it is functionally equivalent but gives the mapping a more meaningful name

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -97,45 +95,46 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    }
-    return candidates;
-  }
-
-  /**
-   * Formats position and its associated candidates' information as {@code Position}. Correlates
-   * one {@code Position} object with one or more {@code DierctoryCandidate}.
+   * Formats a list of positions and their associated candidates' information. Correlates
+   * a {@code Position} object with one or more {@code DierctoryCandidate}. Candidates running
+   * for the same position are assumed to be consecutive in {@code candidates}.
```
kellyktan: nit: Update to match new parameters.

project/src/main/java/com/google/sps/servlets/DataServlet.java
```diff
@@ -97,45 +95,46 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     List<Position> positions = new ArrayList<>(distinctPositions.size());
     for (String positionName : distinctPositions) {
       int startIndex = candidatePositions.indexOf(positionName);
       int endIndex = candidatePositions.lastIndexOf(positionName);
-      positions.add(new Position(positionName, candidates.subList(startIndex, endIndex + 1)));
+      List<DirectoryCandidate> candidates = new ArrayList<>(endIndex - startIndex + 1);
+      for (int i = startIndex; i <= endIndex; i++) {
+        String candidateId = candidateIds.get(i);
+        boolean isIncumbent = candidateIncumbency.get(i);
+        candidates.add(extractCandidateInformation(candidateId, isIncumbent));
```
kellyktan: nit: just do this in one line instead of creating variables.

kellyktan (approved):

PR #9, opened 2020-07-02
Implement the complete structure of news article compilation. Leave small components of skeleton code for future implementation.

KathyFeiyang:

**1. Build the basic workflow for web-scraping and compiling news articles. `WebCrawler` class integrates the workflow.**
1.1 Find news articles URLs from Google Custom Search. [implemented, not yet tested with Custom Search engine]
1.2 Check robots.txt to see if web-scraping is permitted. [implemented]
1.3 Web-scrape the HTML. [implemented]
1.4 Extract textual content from HTML with `ContentExtractor`. [implemented]
1.5 Check content relevancy with `RelevancyChecker`. [implemented, will be extended]
1.6 Process content with `ContentProcessor`. [implemented, will be extended]
1.7 Store processed content in the backend Datastore database. [implemented]
(Parts 1.2, 1.3 and 1.4 will be a basic web-scraping module that can be reused across this project)
- `compileNewsArticle()` of `WebCrawler` is the highest level of abstraction and executes the 7 steps above.

**2. Check robots.txt by using [RobotsTxt](https://github.com/pandzel/RobotsTxt).**
- Implemented crawler delay (may be extended or modified in the future).

**3. Extract textual content from news articles' HTML by using [Apache Tika](https://tika.apache.org/) and [Boilerpipe](https://boilerpipe-web.appspot.com/).**
- The `ContentExtractor` class currently uses `ArticleExtractor` to extract news articles content.
- The `ContentExtractor` class currently extracts two pieces of content:
   (1) Title.
   (2) Main content.

**4. Create Maven Java project.**
- Adjust pom.xml with the relevant dependencies and set up a runnable main for `WebCrawler`.
- To compile and execute `WebCrawler` on a sample URL, run:
```bash
mvn clean compile exec:java
```

**TODOs**
- TODOs are marked in the code with `@TODO [task]`

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [ x ] Tests passing

### DIFF
--- a/web-crawler/README.md
+++ b/web-crawler/README.md
@@ -1,0 +1,7 @@
+This directory contains the web crawler for "Voter Central".
+
+To compile and execute WebCrawler's main, execute this command:
+
+```bash
+mvn clean compile exec:java
+```
--- a/web-crawler/README.md
+++ b/web-crawler/README.md
@@ -1,5 +1,11 @@
 This directory contains the web crawler for "Voter Central".
 
+The use of the Google Cloud Datastore client library requires the project ID to
+be configured:
+```bash
+gcloud config set project <projectId>
+```
+
 To compile and execute WebCrawler's main, execute this command:
 
 ```bash
--- a/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
+++ b/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
@@ -17,7 +17,7 @@ package com.google.sps.data;
 import java.util.Arrays;
 import java.util.List;
 
-/** 
+/**
  * A news article, including its metadata and content.
  */
 public class NewsArticle {
--- a/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
+++ b/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
@@ -22,15 +22,18 @@ import java.util.List;
  */
 public class NewsArticle {
   private String title;
+  private String url;
   private List<String> content;
 
   public NewsArticle() {
     this.title = "";
+    this.url = "";
     this.content = Arrays.asList();
   }
 
-  public NewsArticle(String title, List<String> content) {
+  public NewsArticle(String title, String url, List<String> content) {
     this.title = title;
+    this.url = url;
     this.content = content;
   }
 
@@ -38,7 +41,15 @@ public class NewsArticle {
     return this.title;
   }
 
+  public String getUrl() {
+    return this.url;
+  }
+
   public List<String> getContent() {
     return this.content;
+  }
+
+  public void setContent(List<String> content) {
+    this.content = content;
   }
 }
--- a/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
+++ b/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
@@ -23,12 +23,10 @@ import java.util.List;
 public class NewsArticle {
   private String title;
   private String url;
-  // A list of content blocks together constituting the full content. A content block represents
-  // a paragraph in the news article.
-  private List<String> content;
+  private String content;
   private String abbreviatedContent;
 
-  public NewsArticle(String title, String url, List<String> content) {
+  public NewsArticle(String title, String url, String content) {
     this.title = title;
     this.url = url;
     this.content = content;
@@ -43,7 +41,7 @@ public class NewsArticle {
     return this.url;
   }
 
-  public List<String> getContent() {
+  public String getContent() {
     return this.content;
   }
 
@@ -51,7 +49,7 @@ public class NewsArticle {
     return this.abbreviatedContent;
   }
 
-  public void setContent(List<String> content) {
+  public void setContent(String content) {
     this.content = content;
   }
 
--- a/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
+++ b/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
@@ -23,18 +23,16 @@ import java.util.List;
 public class NewsArticle {
   private String title;
   private String url;
+  // A list of content blocks together constituting the full content. A content block represents
+  // a paragraph in the news article.
   private List<String> content;
-
-  public NewsArticle() {
-    this.title = "";
-    this.url = "";
-    this.content = Arrays.asList();
-  }
+  private String abbreviatedContent;
 
   public NewsArticle(String title, String url, List<String> content) {
     this.title = title;
     this.url = url;
     this.content = content;
+    this.abbreviatedContent = null;
   }
 
   public String getTitle() {
@@ -49,7 +47,15 @@ public class NewsArticle {
     return this.content;
   }
 
+  public String getAbbreviatedContent() {
+    return this.abbreviatedContent;
+  }
+
   public void setContent(List<String> content) {
     this.content = content;
+  }
+
+  public void setAbbreviatedContent(String abbreviatedContent) {
+    this.abbreviatedContent = abbreviatedContent;
   }
 }
--- a/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
+++ b/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
@@ -18,7 +18,7 @@ import java.util.Arrays;
 import java.util.List;
 
 /** 
- * Represents one news article, including its metadata and content.
+ * A news article, including its metadata and content.
  */
 public class NewsArticle {
   private String title;
--- a/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
+++ b/web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
@@ -1,0 +1,44 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+import java.util.Arrays;
+import java.util.List;
+
+/** 
+ * Represents one news article, including its metadata and content.
+ */
+public class NewsArticle {
+  private String title;
+  private List<String> content;
+
+  public NewsArticle() {
+    this.title = "";
+    this.content = Arrays.asList();
+  }
+
+  public NewsArticle(String title, List<String> content) {
+    this.title = title;
+    this.content = content;
+  }
+
+  public String getTitle() {
+    return this.title;
+  }
+
+  public List<String> getContent() {
+    return this.content;
+  }
+}
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
@@ -37,7 +37,7 @@ public class ContentExtractor {
   /** 
    * Extracts textual content from HTML. Packages data into {@code NewsArticle}.
    */
-  public static NewsArticle extractContentFromHtml(InputStream htmlFileStream) {
+  public static NewsArticle extractContentFromHtml(InputStream htmlFileStream, String url) {
     BodyContentHandler bodyHandler = new BodyContentHandler();
     ArticleExtractor articleExtractor = new ArticleExtractor();
     BoilerpipeContentHandler boilerpipeHandler =
@@ -47,7 +47,8 @@ public class ContentExtractor {
     try {
       parser.parse(htmlFileStream, boilerpipeHandler, metadata);
       TextDocument textDocument = boilerpipeHandler.getTextDocument();
-      return formatNewsArticleFromDocument(textDocument);
+      System.out.println(textDocument.getText(false, true));
+      return formatNewsArticleFromDocument(textDocument, url);
     } catch (IOException | SAXException | TikaException e) {
       return new NewsArticle();
     }
@@ -57,13 +58,13 @@ public class ContentExtractor {
    * Formats and packages information of a {@code NewsArticle} from the parsed result of a
    * {@code TextDocument}.
    */
-  private static NewsArticle formatNewsArticleFromDocument(TextDocument textDocument) {
+  private static NewsArticle formatNewsArticleFromDocument(TextDocument textDocument, String url) {
     List<String> content = new LinkedList<>();
     for (TextBlock textBlock : textDocument.getTextBlocks()) {
       if (textBlock.isContent()) {
         content.add(textBlock.getText());
       }
     }
-    return new NewsArticle(textDocument.getTitle(), content);
+    return new NewsArticle(textDocument.getTitle(), url, content);
   }
 }
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
@@ -47,7 +47,6 @@ public class ContentExtractor {
     try {
       parser.parse(htmlFileStream, boilerpipeHandler, metadata);
       TextDocument textDocument = boilerpipeHandler.getTextDocument();
-      System.out.println(textDocument.getText(false, true));
       return formatNewsArticleFromDocument(textDocument, url);
     } catch (IOException | SAXException | TikaException e) {
       return new NewsArticle();
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
@@ -1,0 +1,69 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.webcrawler;
+
+import com.google.sps.data.NewsArticle;
+import de.l3s.boilerpipe.document.TextBlock;
+import de.l3s.boilerpipe.document.TextDocument;
+import de.l3s.boilerpipe.extractors.ArticleExtractor;
+import java.io.InputStream;
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.List;
+import org.apache.tika.exception.TikaException;
+import org.apache.tika.metadata.Metadata;
+import org.apache.tika.parser.html.BoilerpipeContentHandler;
+import org.apache.tika.parser.html.HtmlParser;
+import org.apache.tika.sax.BodyContentHandler;
+import org.xml.sax.SAXException;
+
+/** 
+ * Provides a tool for extracting textual content from HTML.
+ */
+public class ContentExtractor {
+
+  /** 
+   * Extracts textual content from HTML. Packages data into {@code NewsArticle}.
+   */
+  public static NewsArticle extractContentFromHtml(InputStream htmlFileStream) {
+    BodyContentHandler bodyHandler = new BodyContentHandler();
+    ArticleExtractor articleExtractor = new ArticleExtractor();
+    BoilerpipeContentHandler boilerpipeHandler =
+        new BoilerpipeContentHandler(bodyHandler, articleExtractor);
+    HtmlParser parser = new HtmlParser();
+    Metadata metadata = new Metadata();
+    try {
+      parser.parse(htmlFileStream, boilerpipeHandler, metadata);
+      TextDocument textDocument = boilerpipeHandler.getTextDocument();
+      return formatNewsArticleFromDocument(textDocument);
+    } catch (IOException | SAXException | TikaException e) {
+      return new NewsArticle();
+    }
+  }
+
+  /** 
+   * Formats and packages information of a {@code NewsArticle} from the parsed result of a
+   * {@code TextDocument}.
+   */
+  private static NewsArticle formatNewsArticleFromDocument(TextDocument textDocument) {
+    List<String> content = new LinkedList<>();
+    for (TextBlock textBlock : textDocument.getTextBlocks()) {
+      if (textBlock.isContent()) {
+        content.add(textBlock.getText());
+      }
+    }
+    return new NewsArticle(textDocument.getTitle(), content);
+  }
+}
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
@@ -15,14 +15,37 @@
 package com.google.sps.webcrawler;
 
 import com.google.sps.data.NewsArticle;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
 
 /** 
  * Provides a tool for processing textual content.
  */
 public class ContentProcessor {
+  private final static int MAX_WORD_COUNT = 100;
 
-  // @TODO
+  // @TODO [Implement more advanced processing methods.]
+  /** 
+   * Extracts the first {@code MAX_WORD_COUNT} words.
+   */
   public static NewsArticle process(NewsArticle newsArticle) {
+    List<String> rawContent = newsArticle.getContent();
+    List<String> processedContent = new LinkedList<>();
+    int remainingWordCount = MAX_WORD_COUNT;
+    for (String contentBlock : rawContent) {
+      String[] splitContentBlock = contentBlock.split(" ");
+      int wordCount = splitContentBlock.length;
+      int allowedLength = Math.min(wordCount, remainingWordCount);
+      String processedContentBlock =
+          String.join(" ", Arrays.asList(splitContentBlock).subList(0, allowedLength));
+      processedContent.add(processedContentBlock);
+      remainingWordCount -= wordCount;
+      if (remainingWordCount <= 0) {
+        break;
+      }
+    }
+    newsArticle.setContent(processedContent);
     return newsArticle;
   }
 }
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
@@ -1,0 +1,28 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.webcrawler;
+
+import com.google.sps.data.NewsArticle;
+
+/** 
+ * Provides a tool for processing textual content.
+ */
+public class ContentProcessor {
+
+  // @TODO
+  public static NewsArticle process(NewsArticle newsArticle) {
+    return newsArticle;
+  }
+}
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
@@ -14,15 +14,65 @@
 
 package com.google.sps.webcrawler;
 
+import com.google.cloud.language.v1.AnalyzeEntitiesRequest;
+import com.google.cloud.language.v1.AnalyzeEntitiesResponse;
+import com.google.cloud.language.v1.Document;
+import com.google.cloud.language.v1.Document.Type;
+import com.google.cloud.language.v1.EncodingType;
+import com.google.cloud.language.v1.Entity;
+import com.google.cloud.language.v1.EntityMention;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
 import com.google.sps.data.NewsArticle;
 
 /** 
  * Provides a tool for checking the relevancy of content to a candidate.
  */
 public class RelevancyChecker {
+  // @TODO [Calculate a meaningful salience threshold.]
+  private static final double salienceThreshold = 0.5;
 
-  // @TODO
+  /** 
+   * Checks whether the {@code newsArticle} is relevant to the {@code candidateName} of
+   * interest. Defines relevancy as the salience of {@code candidateName} in the content,
+   * and defines sufficient relevancy with {@salienceThreshold}.
+   */
   public static boolean isRelevant(NewsArticle newsArticle, String candidateName) {
-    return true;
+    double totalSalience = 0;
+    int salienceCount = 0;
+    for (String content : newsArticle.getContent()) {
+      try {
+        totalSalience += computeSalienceOfName(content, candidateName);
+        salienceCount++;
+      } catch (Exception e) {
+        continue;
+      }
+    }
+    double averageSalience = salienceCount != 0 ? totalSalience / salienceCount : 0;
+    return averageSalience >= salienceThreshold ? true : false;
+  }
+
+  /** 
+   * Performs entity analysis, and computes the salience score of {@code candidateName} in the
+   * {@code content}. Salience has range [0, 1], with higher salience indicating higher
+   * relevance of {@code candidateName} to {@code content} overall.
+   */
+  private static double computeSalienceOfName(String content, String candidateName)
+      throws Exception {
+    try (LanguageServiceClient language = LanguageServiceClient.create()) {
+      Document doc = Document.newBuilder().setContent(content).setType(Type.PLAIN_TEXT).build();
+      AnalyzeEntitiesRequest request =
+          AnalyzeEntitiesRequest.newBuilder()
+              .setDocument(doc)
+              .setEncodingType(EncodingType.UTF8)
+              .build();
+      AnalyzeEntitiesResponse response = language.analyzeEntities(request);
+      for (Entity entity : response.getEntitiesList()) {
+        if (candidateName.equalsIgnoreCase(entity.getName())) {
+          return entity.getSalience();
+        }
+      }
+      return 0;
+    }
   }
 }
--- a/web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
+++ b/web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
@@ -33,13 +33,20 @@ import java.io.IOException;
 public class RelevancyChecker {
   // @TODO [Calculate a meaningful salience threshold.]
   private static final double SALIENCE_THRESHOLD = 0.5;
-  private static LanguageServiceClient languageServiceClient;
-  static {
-    try {
-      languageServiceClient = LanguageServiceClient.create();
-    } catch (IOException e) {
-      languageServiceClient = null;
-    }
+  private LanguageServiceClient languageServiceClient;
+
+  /** 
+   * Instantiates a {@code LanguageServiceClient} to use the Google Natural Language API.
+   */
+  public RelevancyChecker() throws IOException {
+    this.languageServiceClient = LanguageServiceClient.create();
+  }
+
+  /** 
+   * Initializes a {@code LanguageServiceClient} with the parameter.
+   */
+  public RelevancyChecker(LanguageServiceClient languageServiceClient) {
+    this.languageServiceClient = languageServiceClient;
   }
 
   /** 
@@ -47,11 +54,7 @@ public class RelevancyChecker {
    * interest. Defines relevancy as the salience of {@code candidateName} in the content,
    * and defines sufficient relevancy with {@code SALIENCE_THRESHOLD}.
    */
-  public static boolean isRelevant(NewsArticle newsArticle, String candidateName) throws
-      IOException {
-    if (languageServiceClient == null) {
-      throw new IOException("[ERROR] Failure to create LanguageServiceClient.");
-    }
+  public boolean isRelevant(NewsArticle newsArticle, String candidateName) {
     double salience = computeSalienceOfName(newsArticle.getContent(), candidateName);
     return salience >= SALIENCE_THRESHOLD;
   }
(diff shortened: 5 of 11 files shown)


web-crawler/src/main/java/com/google/sps/data/NewsArticle.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+/** 
+ * Represents one news article, including its metadata and content.
+ */
+public class NewsArticle {
+  private String title;
+  private String url;
+  private List<String> content;
```
kellyktan: Maybe add a comment denoting how the strings are divided up in this `content` list.  (i.e. is each string in the list a paragraph, line, sentence, etc.?)

web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
```diff
@@ -0,0 +1,69 @@
...
+  /** 
+   * Extracts textual content from HTML. Packages data into {@code NewsArticle}.
+   */
+  public static NewsArticle extractContentFromHtml(InputStream htmlFileStream, String url) {
+    BodyContentHandler bodyHandler = new BodyContentHandler();
+    ArticleExtractor articleExtractor = new ArticleExtractor();
+    BoilerpipeContentHandler boilerpipeHandler =
+        new BoilerpipeContentHandler(bodyHandler, articleExtractor);
+    HtmlParser parser = new HtmlParser();
+    Metadata metadata = new Metadata();
```
kellyktan: Can any of these be shared at the class level? (I imagine boilerplatehandler needs to be specific to this method, but maybe the parser could be shared?)

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.google.sps.webcrawler;
+
+import com.google.sps.data.NewsArticle;
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+
+/** 
+ * Provides a tool for processing textual content.
```
kellyktan: Processing how?  Is this the things that "summarizes" articles?

kellyktan: If so, instead of overwriting `NewsArticle.content`, I feel like it might be better to have another variable that holds the "summarized" or "abbreviated" content in case you ever want functionality to expand an article and read the whole thing on the Voter Central site (instead of directing users to the news sites).

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.sps.data.NewsArticle;
+
+// @TODO [Compute the salience of other features, such as location, election, to improve accuracy.]
+/** 
+ * Provides a tool for checking the relevancy of content to a candidate.
+ */
+public class RelevancyChecker {
+  // @TODO [Calculate a meaningful salience threshold.]
+  private static final double salienceThreshold = 0.5;
```
kellyktan: Constant names have different formatting (i.e. `SALIENCE_THRESHOLD`):
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /** 
+   * Performs entity analysis, and computes the salience score of {@code candidateName} in the
+   * {@code content}. Salience has range [0, 1], with higher salience indicating higher
+   * relevance of {@code candidateName} to {@code content} overall.
+   */
+  private static double computeSalienceOfName(String content, String candidateName)
+      throws Exception {
+    try (LanguageServiceClient language = LanguageServiceClient.create()) {
```
kellyktan: Can the LanguageServiceClient be shared at the class level?

web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.tika.metadata.Metadata;
+import org.apache.tika.parser.html.BoilerpipeContentHandler;
+import org.apache.tika.parser.html.HtmlParser;
+import org.apache.tika.sax.BodyContentHandler;
+import org.xml.sax.SAXException;
+
+/** 
+ * Provides a tool for extracting textual content from HTML.
+ */
+public class ContentExtractor {
```
suparnabarua: Rename this class to NewsContentExtractor

web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /** 
+   * Extracts textual content from HTML. Packages data into {@code NewsArticle}.
+   */
+  public static NewsArticle extractContentFromHtml(InputStream htmlFileStream, String url) {
+    BodyContentHandler bodyHandler = new BodyContentHandler();
+    ArticleExtractor articleExtractor = new ArticleExtractor();
+    BoilerpipeContentHandler boilerpipeHandler =
```
suparnabarua: you can do "BoilerpipeContentHandler boilerpipeHandler =
 new BoilerpipeContentHandler(new BodyContentHandler(), new ArticleExtractor())"

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  /** 
+   * Checks robots.txt for permission to web-scrape, scrapes webpage if permitted and extracts
+   * textual content.
+   */
+  public NewsArticle scrapeAndExtractHtml(String url) {
+    String robotsUrl = url.substring(0, url.indexOf("/", url.indexOf("//") + 2) + 1)
+        + "robots.txt";
```
kellyktan: Java has a bunch of URL and URL encoder/decoder libraries built in for extracting portions of the url and building urls.
https://docs.oracle.com/javase/7/docs/api/java/net/URL.html
https://docs.oracle.com/javase/7/docs/api/java/net/URI.html
https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html
https://docs.oracle.com/javase/7/docs/api/java/net/URLDecoder.html

web-crawler/src/main/java/com/google/sps/webcrawler/ContentExtractor.java
```diff
@@ -0,0 +1,69 @@
...
+        new BoilerpipeContentHandler(bodyHandler, articleExtractor);
+    HtmlParser parser = new HtmlParser();
+    Metadata metadata = new Metadata();
+    try {
+      parser.parse(htmlFileStream, boilerpipeHandler, metadata);
+      TextDocument textDocument = boilerpipeHandler.getTextDocument();
+      return formatNewsArticleFromDocument(textDocument, url);
+    } catch (IOException | SAXException | TikaException e) {
```
suparnabarua: You should add to the function documentation that you return a empty newsarticle in case of exceptions (and a brief description of what could cause these exceptions)

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class ContentProcessor {
```
suparnabarua: Rename this to NewsContentProcessor

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+
+  // @TODO [Implement more advanced processing methods.]
+  /** 
+   * Extracts the first {@code MAX_WORD_COUNT} words.
+   */
+  public static NewsArticle process(NewsArticle newsArticle) {
+    List<String> rawContent = newsArticle.getContent();
+    List<String> processedContent = new LinkedList<>();
+    int remainingWordCount = MAX_WORD_COUNT;
+    for (String contentBlock : rawContent) {
```
suparnabarua: nit. add a new line before the for loop

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.cloud.language.v1.EncodingType;
+import com.google.cloud.language.v1.Entity;
+import com.google.cloud.language.v1.EntityMention;
+import com.google.cloud.language.v1.LanguageServiceClient;
+import com.google.cloud.language.v1.LanguageServiceSettings;
+import com.google.sps.data.NewsArticle;
+
+// @TODO [Compute the salience of other features, such as location, election, to improve accuracy.]
+/** 
+ * Provides a tool for checking the relevancy of content to a candidate.
```
suparnabarua: "relevancy of news article content..."

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -0,0 +1,79 @@
...
+    for (String content : newsArticle.getContent()) {
+      try {
+        totalSalience += computeSalienceOfName(content, candidateName);
+        salienceCount++;
+      } catch (Exception e) {
+        continue;
+      }
+    }
+    double averageSalience = salienceCount != 0 ? totalSalience / salienceCount : 0;
+    return averageSalience >= salienceThreshold ? true : false;
```
suparnabarua: you can do "return averageSalience >= salienceThreshold"

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -0,0 +1,79 @@
...
+    int salienceCount = 0;
+    for (String content : newsArticle.getContent()) {
+      try {
+        totalSalience += computeSalienceOfName(content, candidateName);
+        salienceCount++;
+      } catch (Exception e) {
+        continue;
+      }
+    }
+    double averageSalience = salienceCount != 0 ? totalSalience / salienceCount : 0;
```
suparnabarua: This will be cleaner if you do "salienceCount == 0 ? 0 : totalSalience / salienceCount"

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+          nextAccessTimes.replace(robotsUrl,
+                                  System.currentTimeMillis() + grant.getCrawlDelay() * 1000);
+        } else {
+          System.out.println(System.currentTimeMillis());
+          nextAccessTimes.put(robotsUrl,
+                              System.currentTimeMillis() + grant.getCrawlDelay() * 1000);
+        }
+      }
+      // Check permission to access.
+      if (grant == null || grant.hasAccess()) {
```
kellyktan, line 175: Would it be better to check this before you wait for the crawl delay?

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+  public void compileNewsArticle(String candidateName, String candidateId) {
+    List<String> urls = getUrlsFromCustomSearch(candidateName);
+    for (String url : urls) {
+      NewsArticle newsArticle = scrapeAndExtractHtml(url);
+      if (!RelevancyChecker.isRelevant(newsArticle, candidateName)) {
+        continue;
+      }
+      NewsArticle processedNewsArticle = ContentProcessor.process(newsArticle);
+      storeInDatabase(candidateId, processedNewsArticle);
+      // waitForMaxCrawlDelay();
```
suparnabarua: Remove this

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+          }
+        }
+      };
+      String responseBody = httpclient.execute(httpGet, responseHandler);
+      httpclient.close();
+      Gson gson = new Gson();
+      Object jsonResponse = gson.fromJson(responseBody, Object.class);
+      //@TODO [Unpack {@code jsonResponse} and find URLs.]
+      return Arrays.asList(
+          "https://www.cnn.com/2020/06/23/politics/aoc-ny-primary-14th-district/index.html");
```
suparnabarua: You're reusing this. Make this string a constant

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+  }
+
+  /** 
+   * Checks robots.txt for permission to web-scrape, scrapes webpage if permitted and extracts
+   * textual content.
+   */
+  public NewsArticle scrapeAndExtractHtml(String url) {
```
kellyktan: Probably better to return an `Optional<NewsArticle>` instead of returning empty news articles upon failure:
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+      // Check permission to access.
+      if (grant == null || grant.hasAccess()) {
+        InputStream webpageStream = new URL(url).openStream();
+        return ContentExtractor.extractContentFromHtml(webpageStream, url);
+      } else {
+        return new NewsArticle();
+      }
+    } catch (IOException e) {
+      System.out.println("[ERROR] Error occured during web scraping.");
+      return new NewsArticle();
```
suparnabarua: You should document that you return empty news articles in cases of exceptions

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+    try {
+      InputStream robotsTxtStream = new URL(robotsUrl).openStream();
+      RobotsTxt robotsTxt = RobotsTxt.read(robotsTxtStream);
+      String webpagePath = url.substring(url.indexOf("/", url.indexOf("//") + 2));
+      Grant grant = robotsTxt.ask("*", webpagePath);
+      // Check and set crawl delay.
+      if (grant != null && grant.getCrawlDelay() != null) {
+        if (nextAccessTimes.containsKey(robotsUrl)) {
+          // Wait until the crawl delay fully passes.
+          if (System.currentTimeMillis() < nextAccessTimes.get(robotsUrl)) {
```
suparnabarua: I don't like these nested loops. Consider turning all of this into a helper function that does if this && that && those

suparnabarua (changes requested):

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+   * Stores {@code NewsArticle}'s metadata and content into the database, following a predesigned
+   * database schema.
+   * Requires "gcloud config set project project-ID" to be set correctly. Assumes "content" to be
+   * occupy fewer than 1500 bytes.
+   */
+  public void storeInDatabase(String candidateId, NewsArticle newsArticle) {
+    Datastore datastore = DatastoreOptions.getDefaultInstance().getService();
+    Key newsArticleKey = datastore.newKeyFactory()
+        .setKind("NewsArticle")
+        .newKey((new Random()).nextLong());
```
kellyktan: Is this guaranteed to be unique?  Would it be better to do a hash of the url or something?

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,231 @@
...
+    Entity newsArticleEntity = Entity.newBuilder(newsArticleKey)
+        .set("candidateId", datastore.newKeyFactory().setKind("Candidate").newKey(candidateId))
```
kellyktan, line 241: Is it possible for news articles to be relevant to more than one candidate? And if so, it's probably better to only store the article contents once, and have a reference to each relevant candidate.

web-crawler/src/test/java/com/google/sps/WebCrawlerTestor.java
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import com.google.sps.webcrawler.WebCrawler;
+
+public class WebCrawlerTestor {
```
kellyktan: nit: Tester

kellyktan: Also, for usual JUnit test files you usually format it as <class name>Test.  But I imagine this probably has a different purpose.

kellyktan (changes requested):
It'd be great if you could write unit tests for the various classes where you can (when you're able to isolate the logic from the libraries you're using...maybe use mocks?)

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
KathyFeiyang: Got it! I'll specify the processing methods. Right now it's just taking the first 100 words, and later I'll implement summarization. There may be [multiple processing methods](https://docs.google.com/document/d/1J8AfCorIeCkzOax94ubKtfowB2rbVNPsEGP1-vyhKwM/edit#bookmark=id.km2ddfszn8z2) in the future.

KathyFeiyang: Ok, I'll make changes to `NewsArticle`. Thanks! :)

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: From my observation, I think news articles are mostly about one particular candidate. If an article mentions another candidate, it usually only does so briefly. I imagine that the RelevancyChecker would give different salience scores as well.

KathyFeiyang: However, I believe it's definitely necessary to avoid duplication in the database. Right now each news article has a single foreign key pointing to one candidate. I can adjust the data structure to an array of candidate IDs. Or perhaps I can create another table to keep candidate & news article correspondence. In that case I wouldn't have to check for membership in an array.

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
KathyFeiyang: I created a new instance variable `abbreviatedContent` of type String to store the first `MAX_WORD_COUNT` words. The variable `content` would not get modified.

I also modified the database schema to store `content` as a list of text strings. Since the database doesn't allow indexed text strings to exceed 1500 bytes, I excluded these text strings from indexing.

KathyFeiyang: On second thought, I believe using a list of content blocks doesn't add much functionality to the web app. I will still format `content` as a string in the database, but excluded from indexing.

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -24,21 +24,34 @@
...
+  }
 
   /** 
    * Checks whether the {@code newsArticle} is relevant to the {@code candidateName} of
    * interest. Defines relevancy as the salience of {@code candidateName} in the content,
-   * and defines sufficient relevancy with {@salienceThreshold}.
+   * and defines sufficient relevancy with {@code SALIENCE_THRESHOLD}.
    */
-  public static boolean isRelevant(NewsArticle newsArticle, String candidateName) {
+  public static boolean isRelevant(NewsArticle newsArticle, String candidateName) throws
```
kellyktan: I wonder if it'd be better to actually make these non-static, since if it fails to create the `LanguageServiceClient` this whole thing is kind of useless?

You could even have 2 constructors:
- one with 0 params that attempts to create the `LanguageServiceClient` similar to what you have here
- one with a param of `LanguageServiceClient` passed in so that it makes testing easier (i.e. in tests could pass in `MockLanguageServiceClient`)

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -132,65 +144,74 @@ public String handleResponse(final HttpResponse response) throws IOException {
...
-   * textual content.
+   * textual content. Returns an empty {@code Optional<NewsArticle>} in the event of an exception.
    */
-  public NewsArticle scrapeAndExtractHtml(String url) {
-    String robotsUrl = url.substring(0, url.indexOf("/", url.indexOf("//") + 2) + 1)
-        + "robots.txt";
+  public Optional<NewsArticle> scrapeAndExtractHtml(String url) {
     try {
+      URL formattedUrl = new URL(url);
+      String robotsUrl = String.format("%s://%s/robots.txt", formattedUrl.getProtocol(),
```
kellyktan: You could instead use the URL constructor `URL(String protocol, String host, String file)` to get the robots URL object directly.

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -199,23 +220,32 @@ private boolean waitForCrawlDelay(long timeToDelay) {
   /** 
    * Stores {@code NewsArticle}'s metadata and content into the database, following a predesigned
    * database schema.
-   * Requires "gcloud config set project project-ID" to be set correctly. Assumes "content" to be
-   * occupy fewer than 1500 bytes.
+   * Requires "gcloud config set project project-ID" to be set correctly. Data beyond 1500 bytes
+   * be excluded from indexes.
```
kellyktan: Are NewsArticles with more than 1500 bytes of content dropped completely, or is the content just truncated to 1500 bytes?

web-crawler/src/test/java/com/google/sps/WebCrawlerTest.java
```diff
@@ -14,7 +14,7 @@
 
 import com.google.sps.webcrawler.WebCrawler;
 
-public class WebCrawlerTestor {
+public class WebCrawlerTest {
 
   public static void main(String[] args) {
```
kellyktan, line 19: Usually unit tests don't have a main method, but I realize you had different intentions for this class.

kellyktan (changes requested):

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
kellyktan: If this would be rare, then no need to alter your entire database schema to accommodate.  It's probably fine if there are only a few duplicates for certain candidates :)

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
kellyktan: In that case, would it make more sense to have `content` be a single string here as well?

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: `NewsArticle` and `content` are not truncated. It's just that the `content` is not indexed in the database (it also doesn't need to be indexed anyways). Sorry about the confusion. I'll rewrite the comment to make this clear

web-crawler/src/test/java/com/google/sps/WebCrawlerTest.java
KathyFeiyang: I haven't started implementing the unit tests. I'll try to get to those soon

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
KathyFeiyang: The reason why I made it static and used a static initializer is because I originally intended `RelevancyChecker` to be a toolbox class and all its methods to be static. Since static methods access `languageServiceClient`, it has to be static as well.

KathyFeiyang: Would you recommend making `languageServiceClient` and the toolbox method `isRelevant()` non-static? If so, should I also modify the other toolbox classes, like `ContentProcessor`, to make them consistent?

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: Got it! But I'll make sure to keep this in mind if the situation changes

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
KathyFeiyang: Sure, I'll make the changes

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -157,17 +157,17 @@ public String handleResponse(final HttpResponse response) throws IOException {
    */
   public Optional<NewsArticle> scrapeAndExtractHtml(String url) {
     try {
-      URL formattedUrl = new URL(url);
-      String robotsUrl = String.format("%s://%s/robots.txt", formattedUrl.getProtocol(),
-          formattedUrl.getHost());
-      InputStream robotsTxtStream = new URL(robotsUrl).openStream();
+      URL formattedUrlObj = new URL(url);
+      URL robotsUrlObj = new URL(formattedUrlObj.getProtocol(), formattedUrlObj.getHost(),
```
kellyktan: nit: no need for the "Obj" suffixes

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -157,17 +157,17 @@ public String handleResponse(final HttpResponse response) throws IOException {
...
+          "/robots.txt");
```
kellyktan, line 169: Also, I would double check whether the leading `/` is necessary?

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -232,27 +232,20 @@ public void storeInDatabase(String candidateId, NewsArticle newsArticle) {
...
         .set("url", newsArticle.getUrl())
-        .set("content", excludeStringFromIndexes(combineContentInHtml(newsArticle.getContent())))
+        .set("content", excludeStringFromIndexes(newsArticle.getContent()))
         .set("abbreviatedContent", excludeStringFromIndexes(newsArticle.getAbbreviatedContent()))
         .build();
     datastore.put(newsArticleEntity);
   }
 
   /** 
    * Converts {@code String} to {@code StringValue} and excludes the data from indexes, to avoid
```
kellyktan, line 251: I would elaborate on how this is specific to DataStore functionality (i.e. what does "indexes" mean in this context?)

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
kellyktan: Yes, so `languageServiceClient` would be an instance variable, and both `isRelevant()` and `computeSalienceOfName()` would be non-static methods.

For the other toolbox classes, probably no need to make them non-static unless they are inconvenient for testing (i.e. I could see maybe making `NewsContentExtractor` non-static if you wanted to use a mock `HtmlParser` in tests). I mostly mentioned this here since there could be errors upon creation of `LanguageServiceClient`, so you'd probably just want to catch that upfront in the initial class instantiation.

kellyktan (changes requested):

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: Yes, otherwise `robotsUrlObj.toString()` doesn't return the correct URL string.

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: Got it, thank you! I originally wanted to make this variable distinct from String-typed URLs. I'll remove it

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: Will do!

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
kellyktan: For instance, when you use the `RelevancyChecker` in `WebCrawler`, the `isRelevant()` call in `compileNewsArticle()` will always fail if the initial creation of `LanguageServiceClient` failed, so there's no real point to keep trying to call `isRelevant()` over and over again since you know it'll always fail at that point.

KathyFeiyang: This makes sense, I agree that this is much more important than whether the toolbox class is static. I'll make the changes

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
kellyktan: Cool, thanks for confirming!

web-crawler/src/main/java/com/google/sps/webcrawler/ContentProcessor.java
KathyFeiyang: This change makes the code significantly more succinct.

web-crawler/README.md
```diff
@@ -1,5 +1,11 @@
 This directory contains the web crawler for "Voter Central".
 
+The use of the Google Cloud Datastore client library requires the project ID to
+be configured:
+```bash
+gcloud config set project <projectId>
```
kellyktan, line 6: Oh, do you need me to set up a new GCP project?

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -33,25 +33,28 @@
 public class RelevancyChecker {
```
kellyktan, line 34: I just remembered the stuff about best practices for class JavaDocs: https://docs.google.com/document/d/17zPdB6a-rw5ZBvm-XvZcxv9AyzMYsSHy0B3U1pVA0s8/edit#heading=h.3bwgzxnein2u

They have differences between instance classes and static classes.

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -33,25 +33,28 @@
...
+
+  /** 
+   * Instantiates a {@code LanguageServiceClient} to use the Google Natural Language API.
+   */
+  public RelevancyChecker() throws IOException {
+    this.languageServiceClient = LanguageServiceClient.create();
+  }
+
+  /** 
+   * Initializes a {@code LanguageServiceClient} with the parameter.
```
kellyktan: nit: Constructor JavaDoc best practices:
https://docs.google.com/document/d/17zPdB6a-rw5ZBvm-XvZcxv9AyzMYsSHy0B3U1pVA0s8/edit#heading=h.wupz364e1vve

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -80,11 +80,13 @@ public void compileNewsArticle(String candidateName, String candidateId) {
...
+      RelevancyChecker relevancyChecker;
       try {
-        if (!RelevancyChecker.isRelevant(newsArticle, candidateName)) {
-          continue;
-        }
-      } catch (Exception e) {
+        relevancyChecker = new RelevancyChecker();
+      } catch (IOException e) {
+        continue;
+      }
```
kellyktan: Can you instantiate this outside of the for loop, once for this function? (or better yet, make it an instance variable so that WebCrawler doesn't keep creating this object)

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -157,17 +159,17 @@ public String handleResponse(final HttpResponse response) throws IOException {
    */
   public Optional<NewsArticle> scrapeAndExtractHtml(String url) {
     try {
-      URL formattedUrlObj = new URL(url);
-      URL robotsUrlObj = new URL(formattedUrlObj.getProtocol(), formattedUrlObj.getHost(),
+      URL formattedUrl = new URL(url);
```
kellyktan: What makes this a "formattedUrl"?  Maybe just pass in the URL object as the function param instead?

kellyktan (changes requested):

web-crawler/README.md
KathyFeiyang: No I don't think so. I've using the existing one and it works well!

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
KathyFeiyang: Thanks for pointing this out! I'll read through the doc and make changes

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -28,23 +28,24 @@
 
 // @TODO [Compute the salience of other features, such as location, election, to improve accuracy.]
 /**
- * Provides a tool for checking the relevancy of news article content to a candidate.
+ * A {@code LanguageServiceClient} that performs entity analysis to check the relevancy of news
```
kellyktan: Did you mean `{@code RelevancyChecker}`? (here and below?)

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -50,16 +50,27 @@
...
   private final static String CUSTOM_SEARCH_ENGINE_ID = "";
   private final static String TEST_URL =
       "https://www.cnn.com/2020/06/23/politics/aoc-ny-primary-14th-district/index.html";
+  private RelevancyChecker relevancyChecker;
+  // Mappings of (website robots.txt URL, the next allowed time to access, in milliseconds) for
+  // respecting the required crawl delay.
+  private Map<String, Long> nextAccessTimes = new HashMap<>();
+
+  /** 
+   * Constructs a {@code RelevancyChecker} instance.
```
kellyktan: Constructs a WebCrawler

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -253,7 +261,14 @@ private StringValue excludeStringFromIndexes(String content) {
 
   // For testing purposes.
   public static void main(String[] args) {
-    WebCrawler myWebCrawler = new WebCrawler();
+    WebCrawler myWebCrawler;
+    try {
+      myWebCrawler = new WebCrawler();
+    } catch (IOException e) {
+      System.out.println("[ERROR] " + e);
+      System.exit(0);
```
kellyktan: Doesn't 0 usually everything succeeded?

web-crawler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
KathyFeiyang: Yeah I think RelevancyChecker makes more sense. I'll make the changes

KathyFeiyang: On line 42, I still kept the comment as `LanguageServiceClient` because the exception would come from that

kellyktan (approved):

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+      };
+      String responseBody = httpclient.execute(httpGet, responseHandler);
+      httpclient.close();
+      Gson gson = new Gson();
+      Object jsonResponse = gson.fromJson(responseBody, Object.class);
+      //@TODO [Unpack {@code jsonResponse} and find URLs.]
+    } catch (IOException e){
+      System.out.println("[ERROR] Error occurred with fetching URLs from Custom Search: " + e);
+      // For testing purposes (commented off).
```
suparnabarua: remove these comments

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,274 @@
...
+  /** 
+   * Searches for {@code candidateName} in the Google Custom Search engine and finds URLs of news
+   * articles. Returns an empty list if no valid URLs are found.
+   *
+   * @see <a href=
+   *    "https://hc.apache.org/httpcomponents-client-ga/httpclient/examples/org/apache/" +
+   *    "http/examples/client/ClientWithResponseHandler.java">Code reference</a>
+   */
+  public List<URL> getUrlsFromCustomSearch(String candidateName) {
+    // For testing purposes.
```
suparnabarua: You have "For testing purposes" all over this function. You should make it clear whether this function is meant only to be used in tests OR if it's for hardcoded purposes

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -0,0 +1,274 @@
...
+      URL robotsUrl = new URL(url.getProtocol(), url.getHost(),
+          "/robots.txt");
+      InputStream robotsTxtStream = robotsUrl.openStream();
+      RobotsTxt robotsTxt = RobotsTxt.read(robotsTxtStream);
+      String webpagePath = url.getPath();
+      Grant grant = robotsTxt.ask("*", webpagePath);
+      // Check permission to access and respect the required crawl delay.
+      if (grant == null || grant.hasAccess()) {
+        if (grant != null && grant.getCrawlDelay() != null) {
+          if (!waitForAndSetCrawlDelay(grant, robotsUrl.toString())) {
```
suparnabarua: You can combine this check with the checks from line 177

suparnabarua (changes requested):

web-crawler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: Apologies for the confusion! This is because I haven't been given access to a custom search engine yet, so the function will throw errors. However, I still wanted the web crawler to be tested from the beginning to the end, so I incorporated this function. In order to do so, I have to add hard-coded URLs, for the next function to have URLs to scrape.

KathyFeiyang: I added descriptions for the above.

tjyoung99 (approved):

ganewman (approved):
LGTM pending resolution of other comments

suparnabarua (approved):

PR #14, opened 2020-07-10
Write unit tests for the news articles compilation process using JUnit and Mockito.

KathyFeiyang:

**_Write unit tests for the news articles compilation process using JUnit and Mockito._**

**1. The unit tests cover the following parts:**
1.1 `WebCrawler`: checking robots.txt
1.2 `WebCrawler`: scraping HTML
1.3 `ContentExtractor`: extracting textual content
1.4 `RelevancyChecker`: checking content relevancy
1.5 `ContentProcessor`: processing content.
1.6 `WebCrawler`: constructing Datastore entities

The unit tests for `WebCrawler`: getting URLs from Google Custom Search are yet to be written, pending getting access to the service.

**2. The unit tests can be run via:**
```bash
cd voter-central/info-compiler/
mvn test
```
**3. Considerations and potential future development are documented in the comments.**

### Is this fixing an issue?

none

none

- [ x ] Test written/updating
- [ x ] Tests passing


info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -67,6 +70,15 @@
    *     of permission to access required libraries.
    */
   public WebCrawler() throws IOException {
+    this.datastore = DatastoreOptions.getDefaultInstance().getService();
+    this.relevancyChecker = new RelevancyChecker();
```
kellyktan:
```suggestion
    Datastore datastore = DatastoreOptions.getDefaultInstance().getService();
    this(datastore);
```
Can chain constructors to avoid duplication.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,133 @@
...
+          "primary election on Tuesday for New York's 14th Congressional District, CNN projects.");
+  private InputStream validWebpageStream;
+
+  @Before
+  public void createInputStream() throws IOException {
+    validWebpageStream = new URL(URL).openStream();
+  }
+
+  @Test
+  public void extractContentFromValidWebpage() {
```
kellyktan: nit: use JUnit test method naming conventions (i.e. `extractContentFromHtml_validWebpage()`)
https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names

kellyktan (changes requested):

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -193,9 +227,9 @@ public String handleResponse(final HttpResponse response) throws IOException {
   /**
    * Waits for the required crawl delay to pass if necessary and makes a note of the required
    * crawl delay. Returns true if the aforementioned process succeeded. {@code grant} is expected
-   * to non-null.
+   * to non-null. This method is made public for testing purposes.
    */
-  private boolean waitForAndSetCrawlDelay(Grant grant, String url) {
+  public boolean waitForAndSetCrawlDelay(Grant grant, String url) {
```
kellyktan: I think  you should be able to just make these methods "protected" instead of public if you wanted, since the tests fall in the same package, technically

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,133 @@
...
+    Assert.assertTrue(potentialNewsArticle.isPresent());
+    NewsArticle newsArticle = potentialNewsArticle.get();
+    Assert.assertEquals(newsArticle.getTitle(), NEWS_ARTICLE.getTitle());
+    Assert.assertEquals(newsArticle.getUrl(), NEWS_ARTICLE.getUrl());
+    Assert.assertTrue(newsArticle.getContent().contains(NEWS_ARTICLE.getContent()));
+    Assert.assertNull(newsArticle.getAbbreviatedContent());
+  }
+
+  @Test
+  public void eextractContentFromHtml_validWebpageWithWrongUrlParam() {
```
kellyktan: extra e?

kellyktan (changes requested):

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
kellyktan: Oh sorry, I think I actually meant default (unless you actually need the additional functionality of "protected"):
https://www.tutorialspoint.com/what-are-the-differences-between-protected-and-default-access-specifiers-in-java#:~:text=What%20are%20the%20differences%20between%20protected%20and%20default%20access%20specifiers%20in%20Java%3F,-JavaObject%20Oriented&text=The%20Protected%20access%20specifier%20is,visible%20in%20the%20same%20package.

This also applies to variables as well

KathyFeiyang: Ah ok! Thank you

info-compiler/src/main/java/com/google/sps/webcrawler/NewsContentExtractor.java
```diff
@@ -32,7 +32,7 @@
 
 /** Static utilities for extracting textual content from HTML pages. */
 public class NewsContentExtractor {
-  private static HtmlParser parser = new HtmlParser();
+  private static final HtmlParser parser = new HtmlParser();
```
kellyktan: nit: PARSER
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

info-compiler/src/main/java/com/google/sps/webcrawler/NewsContentProcessor.java
```diff
@@ -19,7 +19,7 @@
 
 /** Static utilities for processing textual content, such as abbreviation. */
 public class NewsContentProcessor {
-  public final static int MAX_WORD_COUNT = 100;
+  default static final int MAX_WORD_COUNT = 100;
```
kellyktan: *Ah sorry you can leave off the word "default" for these (methods & variables) and it should have the same level of visibility.
```suggestion
  static final int MAX_WORD_COUNT = 100;
```

kellyktan (changes requested):

info-compiler/src/main/java/com/google/sps/webcrawler/NewsContentProcessor.java
kellyktan: Oops looks like you did this already.

KathyFeiyang: Thanks for the reminder! I just realized too and removed those

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -64,6 +67,15 @@
    *     of permission to access required libraries.
    */
   public WebCrawler() throws IOException {
+    this(DatastoreOptions.getDefaultInstance().getService());
+    this.relevancyChecker = new RelevancyChecker();
```
kellyktan: this shouldn't be necessary since the other constructor constructs a relevancy checker as well.

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -157,7 +169,7 @@ public String handleResponse(final HttpResponse response) throws IOException {
     return urls;
   }
 
-  /**
+  /**
```
kellyktan: Extra space?

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -186,12 +198,37 @@ public String handleResponse(final HttpResponse response) throws IOException {
     }
   }
 
+  /** 
+   * For testing purposes.
+   */
+  Optional<NewsArticle> scrapeAndExtractHtml(URL url, Grant grant) {
```
kellyktan: I would refactor the other scrapeAndExtractHtml to use as much of this as possible since this is the function actually being tested.

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
```diff
@@ -246,6 +283,20 @@ public void storeInDatabase(String candidateId, NewsArticle newsArticle) {
     datastore.put(newsArticleEntity);
   }
 
+  /**
+   * For testing purposes.
+   */
+  FullEntity storeInDatabase(String candidateId, NewsArticle newsArticle,
```
kellyktan: Same here, rewrite above to reuse the tested logic.

kellyktan (changes requested):

info-compiler/src/main/java/com/google/sps/webcrawler/WebCrawler.java
KathyFeiyang: I've made the switch from Mockito mocks of Datastore to Datastore emulators. The extra testable method is removed. Now I'm directly testing on the real `storeInDatabase()`.

info-compiler/src/test/java/com/google/sps/RelevancyCheckerTest.java
```diff
@@ -0,0 +1,167 @@
...
+  @Before
+  public void initialize() throws IOException {
+    mockRelevancyChecker = mock(RelevancyChecker.class);
+    LanguageServiceClient languageServiceClient = mock(LanguageServiceClient.class);
+    relevancyChecker = new RelevancyChecker(languageServiceClient);
+    realRelevancyChecker = new RelevancyChecker();
+  }
+
+  @Test
+  public void computeSalienceOfName_checkWithMockComputationThatFindsEverythingRelevant() {
```
suparnabarua: Rename this to computeSalienceOfName_findsEverythingRelevant
You don't need to mention in the test name that you're using mocks as that is generally expected

info-compiler/src/test/java/com/google/sps/RelevancyCheckerTest.java
```diff
@@ -0,0 +1,167 @@
...
+    // Check content relevancy with a mock relevancy checker that always computes a salience
+    // score higher than {@code SALIENCE_THRESHOLD}. The mock will determine relevancy by
```
suparnabarua: You don't need to write the comments after the first sentence. It is expected that you set up the mock to behave as the real service you're mocking, so you don't need to explicitly mention that

info-compiler/src/test/java/com/google/sps/RelevancyCheckerTest.java
```diff
@@ -0,0 +1,167 @@
...
+    when(mockRelevancyChecker.isRelevant(anyObject(), anyString()))
+        .thenCallRealMethod();
+    Assert.assertTrue(mockRelevancyChecker.isRelevant(relevantNewsArticle, CANDIDATE_NAME));
+    Assert.assertTrue(mockRelevancyChecker.isRelevant(irrelevantNewsArticle, CANDIDATE_NAME));
+  }
+
+  @Test
+  public void computeSalienceOfName_checkWithMockComputationThatFindsEverythingIrrelevant() {
+    // Check content relevancy with a mock relevancy checker that always computes a salience
+    // score lower than {@code SALIENCE_THRESHOLD}. The mock will determine relevancy by
```
suparnabarua: Same comment as above test

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -41,10 +41,13 @@
...
   public RelevancyChecker() throws IOException {
-    this.languageServiceClient = LanguageServiceClient.create();
+    this(LanguageServiceClient.create());
   }
 
-  public RelevancyChecker(LanguageServiceClient languageServiceClient) {
+  /**
+   * For testing purposes.
+   */
+  RelevancyChecker(LanguageServiceClient languageServiceClient) {
```
suparnabarua, line 48: You shouldn't be creating methods just for testing purposes .Instead,  redesign this class so that the constructor takes a languageServiceClient object. If the object is null, you can call LanguageServiceClient.create(), otherwise use the one that's passed in

suparnabarua (changes requested):

info-compiler/src/test/java/com/google/sps/WebCrawlerTest.java
```diff
@@ -59,13 +62,15 @@
...
-    this.datastore = mock(Datastore.class);
-    this.webCrawler = new WebCrawler(datastore);
+  @BeforeClass
+  public static void initialize() throws InterruptedException, IOException {
+    datastoreHelper = LocalDatastoreHelper.create();
+    datastoreHelper.start();
+    Datastore datastore = datastoreHelper.getOptions().getService();
+    webCrawler = new WebCrawler(datastore);
     NEWS_ARTICLE.setAbbreviatedContent(ABBREVIATED_CONTENT);
   }
```
kellyktan, line 90: I think it's fine to keep these non-static and just create a new LocalDatastoreHelper for each test case (i.e. in `@Before` instead of `@BeforeClass`), and then you don't have to reset the datastore for each test.

kellyktan (changes requested):

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
KathyFeiyang: Got it!
To confirm: For Java, other instantiations of RelevancyChecker must pass in null explicitly: `new RelevancyChecker(null)` Would this be desirable?

kellyktan: @suparnabarua  I think it's fine to have duplicate constructors, with one for testing.  I've definitely done this in my work at Google.

info-compiler/src/test/java/com/google/sps/WebCrawlerTest.java
KathyFeiyang: I just implemented this, but `datastoreHelper.stop()` sometimes generates a `java.net.ConnectException: Connection refused (Connection refused)`. This connection problem results from the Datastore emulator making HTTP requests. The number of times this problem occurs also varies between different runs of `WebCrawlerTest`. I believe it may be better to limit the number of times DatastoreHelper is created/destroyed to the minimum amount.

kellyktan: Okay that makes sense.  In that case, would it make sense to move the "reset datastore" to a `@Before` method instead?

KathyFeiyang: Yeah definitely! That seems much better. I'll make the change

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
suparnabarua: as per offline discussion, keep it as it is

KathyFeiyang: got it! thank you so much!

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,133 @@
...
+          "CNNPolitics",
+          URL,
+          "Washington (CNN)Freshman Democratic Rep. Alexandria Ocasio-Cortez will defeat former " +
+          "longtime CNBC correspondent and anchor Michelle Caruso-Cabrera in a Democratic " +
+          "primary election on Tuesday for New York's 14th Congressional District, CNN projects.");
+  private InputStream validWebpageStream;
+
+  @Before
+  public void createInputStream() throws IOException {
+    validWebpageStream = new URL(URL).openStream();
```
kellyktan: Do you need to close this ever?

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.xml.sax.SAXException;
+import static org.mockito.Mockito.*;
+
+/**
+ * A tester for news article content extraction.
+ */
+@RunWith(JUnit4.class)
+public final class NewsContentExtractorTest {
+  private final static String URL =
+    "https://www.cnn.com/2020/06/23/politics/aoc-ny-primary-14th-district/index.html";
```
kellyktan, line 45: Ideally you shouldn't be depending on anything "real" in these tests (i.e. what happens if this link is moved or deleted?).  Is it possible to mock the InputStream to have it return "expected" content, not the actual webpage contents?

kellyktan (changes requested):

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: Thanks this makes sense! I will try

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: Oh I missed that. I looked it up and I should close it

info-compiler/src/test/java/com/google/sps/WebCrawlerTest.java
```diff
@@ -12,12 +12,256 @@
...
+    // {@code NewsContentExtractor} to aid test-driven development.
+    URL url = new URL(VALID_URL);
+    URL robotsUrl = new URL(url.getProtocol(), url.getHost(), "/robots.txt");
+    Grant grant = mock(Grant.class);
+    when(grant.hasAccess()).thenReturn(true);
+    when(grant.getCrawlDelay()).thenReturn(DELAY);
+    WebCrawler webCrawlerSpy = spy(webCrawler);
+    Optional<NewsArticle> potentialNewsArticle = webCrawler.politelyScrapeAndExtractFromHtml(
+        grant, robotsUrl, url);
+    //verify(webCrawlerSpy, times(1)).waitForAndSetCrawlDelay(anyObject(), anyString());
```
suparnabarua: remove the comment

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: I created a new test `extractContentFromHtml_alwaysValidWebpage()` that has a mock `InputStream` returning predefined content.
The content extractor repeatedly calls [`read(byte[], int, int)`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A-int-int-) on the `InputStream`. So, I mocked the object to return -1 for `read()`, which is the indicator of the end of the stream. Then the extractor will parse and extract the empty content.
Do you think I should write sample HTML for the extractor and convert the HTML to `InputStream`? It's definitely doable, thought it will take some tweaking to define the "expected" content (since the implementation of the extractor is hidden from me).

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Test
+  public void extractContentFromHtml_validWebpage() {
+    // Extract content and meta data from a valid webpage. The extracted information should be
+    // consistent with {@code URL} and {@code NEWS_ARTICLE}. Content processing hasn't occurred
+    // so the abbreviated content is null.
+    Optional<NewsArticle> potentialNewsArticle =
+        NewsContentExtractor.extractContentFromHtml(validWebpageStream, URL);
+    Assert.assertTrue(potentialNewsArticle.isPresent());
+    NewsArticle newsArticle = potentialNewsArticle.get();
```
suparnabarua: One thing I want to point out: If you had implemented an equals() method in NewsArticle you could just build a "expectedNewsArticle" object and assert that the two objects are equal instead of asserting the validity of each field. This could come in handy the more tests you write (and for future classes as well). It's up to you whether you want to implement this.

suparnabarua (changes requested):

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: Got it! I think this will be beneficial for the future and make the code cleaner, since I won't have to explicitly write what fields there are and how many fields there are. I'll make the adjustment.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
kellyktan: Yeah actually I would just mock the HtmlParser so that whatever it returns is fully controllable by you in these tests.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: This is done!

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: Oh gotcha! That works too. Will do

KathyFeiyang: I'll make the class non-static first like `RelevancyChecker`

KathyFeiyang: I wrote `extractContentFromHtml_emptyDocParser()` that has a mock parser. The tricky part is that the parser doesn't directly return any results. The code looks like this;
```
      parser.parse(htmlFileStream, boilerpipeHandler, metadata);
      TextDocument textDocument = boilerpipeHandler.getTextDocument();
```
`boilerpipeHandler`, which is the input parameter, gets modified in `parser.parse()` and then returns `getTextDocument()`. By a invocation lambda function, I can get the reference to `boilerpipeHandler`. However the handler class doesn't seem to have a setter method that allows me to set its content, and I'm also not allowed to replace the original reference to point to a mock handler object.
Because of this situation, this method appears to be the best fit: 
[`endDocument()`](https://docs.oracle.com/javase/7/docs/api/org/xml/sax/ContentHandler.html#endDocument()). I called this method on `boilerpipeHandler` to indicate that the document has ended. As a result, the extracted content would predictably be empty. The constructed `NewsArticle` would be the same as `NEWS_ARTICLE_WITH_EMPTY_CONTENT`.

KathyFeiyang: Let me know if this is okay, or if there is a better way

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -47,8 +49,19 @@
...
+          "CNNPolitics",
+          WRONG_URL,
+          "Washington (CNN)Freshman Democratic Rep. Alexandria Ocasio-Cortez will defeat former " +
+          "longtime CNBC correspondent and anchor Michelle Caruso-Cabrera in a Democratic " +
+          "primary election on Tuesday for New York's 14th Congressional District, CNN projects.");
+  private final static NewsArticle NEWS_ARTICLE_WITH_EMPTY_CONTENT =
+      new NewsArticle(EMPTY, URL, EMPTY);
   private final static int END_OF_STREAM_INDICATOR = -1;
   private InputStream realWebpageStream;
+  private NewsContentExtractor realNewsContentExtractor = new NewsContentExtractor();
```
kellyktan, line 59: Now remove tests using the "real" news.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -77,33 +87,51 @@ public void extractContentFromHtml_realValidWebpageWithWrongUrlParam() {
...
+    // receive an end-of-document notification via {@code endDocument()}. Therefore, an empty
+    // article will be extracted, while the URL is set correctly to the passed in parameter.
+    // Content processing hasn't occurred so the abbreivated content is null.
+    HtmlParser parser = mock(HtmlParser.class);
+    doAnswer(parseInvocation -> {
+               BoilerpipeContentHandler boilerpipeHandler =
+                   (BoilerpipeContentHandler) parseInvocation.getArguments()[1];
+               boilerpipeHandler.endDocument();
+               return null;
+            }).when(parser).parse(anyObject(), anyObject(), anyObject());
```
kellyktan: And write more tests that do this, where you set up the boilerpipehandler to contain the TextDocument for the test case.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -77,33 +87,51 @@ public void extractContentFromHtml_realValidWebpageWithWrongUrlParam() {
...
-    // set correctly to {@code URL}. Content processing hasn't occurred so the abbreivated content
-    // is null.
-    InputStream alwaysValidStream = mock(InputStream.class);
-    when(alwaysValidStream.read(anyObject(), anyInt(), anyInt()))
+  public void extractContentFromHtml_alwaysValidAndEmptyWebpageStream() throws IOException {
+    // Extract content and meta data from an always valid and empty webpage that contains no
+    // content. The extraction process is correct, and extracted title and content are {@code
+    // EMPTY}. URL is set correctly to {@code URL}. Content processing hasn't occurred so the
+    // abbreivated content is null.
+    InputStream alwaysValidAndEmptyStream = mock(InputStream.class);
```
kellyktan: No need to mock the InputStream anymore if you're mocking the parser anyways (and can return the textdocument you want)

kellyktan (changes requested):

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
kellyktan: Looks like there's an "addTextBlock" inherited method? https://tika.apache.org/1.0/api/org/apache/tika/parser/html/BoilerpipeContentHandler.html

kellyktan: and "setTitle"

kellyktan: oh sad apparently "addTextBlock" is protected :/ I guess probably the only other way I can think to get around this is to overload extractContentFromHtml and have another version that also takes in a BoilerpipeContentHandler so you can mock it to return whatever TextDocument you want?

kellyktan: sorry this got so complicated 😬

KathyFeiyang: No this was a good dive into Mockito!
Then I'll write another version of the method

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,187 @@
...
+ * A tester for news article content extraction.
+ */
+@RunWith(JUnit4.class)
+public final class NewsContentExtractorTest {
+  private final static String URL =
+    "https://www.cnn.com/2020/06/23/politics/aoc-ny-primary-14th-district/index.html";
+  private final static String WRONG_URL =
+    "https://www.wrong.com";
+  private final static String EMPTY = "";
+  private final static NewsArticle NEWS_ARTICLE =
```
suparnabarua: nit. rename this to VALID_NEWS_ARTICLE or EXPECTED_NEWS_ARTICLE

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
suparnabarua: Looks great, thanks! One nit:you can change the assert to:
Assert.assertEquals(EXPECTED_NEWS_ARTICLE, potentialNewsArticle.get()).
If you look at the Assert Javadoc, the expected argument comes first. I would make that change for the rest of your tests as well.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: I removed all tests that read from real URLs. I kept `realNewsContentExtractor` because it is used to test bad input streams that throw exceptions.

info-compiler/src/main/java/com/google/sps/data/NewsArticle.java
```diff
@@ -51,4 +58,27 @@ public void setContent(String content) {
...
+   * because, when scraping real webpages, the scraped content also includes the actual news
+   * articles body as well as changing content such as ads, descriptions of other articles.
+   */
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof NewsArticle)) {
+      return false;
+    }
+    NewsArticle other = (NewsArticle) obj;
+    if (other == this) {
```
suparnabarua: I would do the object comparison first and then check the instanceof. If obj is already == this, you're wasting a step by casting it to NewsArticle.

suparnabarua (changes requested):

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: Will do!

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -41,57 +46,83 @@
...
+  private final static TextDocument TEXT_DOC_WITHOUT_TITLE =
+      new TextDocument(Arrays.asList(TEXT_BLOCK));
+  private final static NewsArticle NEWS_ARTICLE_REGULAR =
+      new NewsArticle(TITLE, URL, CONTENT);
+  private final static NewsArticle NEWS_ARTICLE_REGULAR_WRONG_URL =
+      new NewsArticle(TITLE, WRONG_URL, CONTENT);
+  private final static NewsArticle NEWS_ARTICLE_WITHOUT_TITLE =
+      new NewsArticle(EMPTY, URL, CONTENT);
+  private final static NewsArticle NEWS_ARTICLE_WITH_EMPTY_TITLE_CONTENT =
       new NewsArticle(EMPTY, URL, EMPTY);
```
kellyktan: Instead of constructing all of these variations ahead of time, I would just do something like this for the tests:
`Assert.assertEquals(new NewsArticle(expected TITLE, expected URL, expected CONTENT), potentialNewsArticle.get());`

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -41,57 +46,83 @@
...
-  private final static int END_OF_STREAM_INDICATOR = -1;
-  private InputStream realWebpageStream;
+  private InputStream webpageStream;
   private NewsContentExtractor realNewsContentExtractor = new NewsContentExtractor();
 
   @Before
   public void createInputStream() throws IOException {
-    realWebpageStream = new URL(URL).openStream();
+    webpageStream = new URL(URL).openStream();
```
kellyktan: Do you need this to be a real inputstream? Probably can just have a mock object instead of opening a sream.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -0,0 +1,200 @@
...
+  // @Test
+  // public void extractContentFromHtml_SAXExceptionStream() throws IOException {
+  //   // The webpage stream throws {@code SAXException} when being read. This exception should be
+  //   // caught and no content will be extracted.
+  //   InputStream badStream = mock(InputStream.class);
+  //   when(badStream.read(anyObject(), anyInt(), anyInt())).thenThrow(new SAXException());
+  //   Optional<NewsArticle> potentialNewsArticle =
+  //       realNewsContentExtractor.extractContentFromHtml(badStream, URL);
+  //   Assert.assertFalse(potentialNewsArticle.isPresent());
+  // }
```
kellyktan: Is there a reason there are so many commented-out tests (here and in other files)?

info-compiler/src/test/java/com/google/sps/NewsContentProcessorTest.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Test
+  public void process_longContent() {
+    // Process {@code LONG_CONTENT} and extract abbreviated content as {@code MAX_CONTENT}, which
+    // contains the {@code MAX_WORD_COUNT} of words. After content processing, the title, URL and
+    // content remain the same.
+    NewsArticle newsArticle = new NewsArticle(TITLE, URL, LONG_CONTENT);
+    NewsArticle processedNewsArticle = NewsContentProcessor.process(newsArticle);
+    Assert.assertEquals(MAX_CONTENT_NEWS_ARTICLE, processedNewsArticle);
```
kellyktan: Similar to my other comments, instead of constructing the expected article outside of the test, I'd just add a line here like:
```suggestion
    NewsArticle expectedArticle = new NewsArticle(newsArticle);
    expectedArticle.setAbbreviatedContent(MAX_CONTENT);
    Assert.assertEquals(expectedArticle, processedNewsArticle);
```

kellyktan (changes requested):

info-compiler/src/test/java/com/google/sps/NewsContentProcessorTest.java
kellyktan: *Or just `NewsArticle expectedArticle = new NewsArticle(TITLE, URL, LONG_CONTENT, ABBREVIATED_CONTENT);`

suparnabarua (approved):

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: Some are kinda left as TODOs. Some require additional Mockito configuration (for final classes), and the others are throwing errors that may or may not be resolved. I'll remove those that appear unsolvable.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: That makes sense! I'll mock it

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -65,15 +65,15 @@
   private NewsContentExtractor realNewsContentExtractor = new NewsContentExtractor();
 
   @Before
-  public void createInputStream() throws IOException {
-    webpageStream = new URL(URL).openStream();
+  public void createInputStream() {
+    webpageStream = mock(InputStream.class);
```
kellyktan, line 65: maybe move mock parser and boilerpipehander into the `@Before` as well.  Do you still need the `@After`?

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -144,7 +142,7 @@ public void extractContentFromHtml_emptyDocParser() throws IOException, SAXExcep
         newsContentExtractor.extractContentFromHtml(
             webpageStream, NEWS_ARTICLE_WITH_EMPTY_TITLE_CONTENT.getUrl());
     Assert.assertTrue(potentialNewsArticle.isPresent());
-    Assert.assertEquals(NEWS_ARTICLE_WITH_EMPTY_TITLE_CONTENT, potentialNewsArticle.get());
+    Assert.assertEquals(new NewsArticle(EMPTY, URL, EMPTY), potentialNewsArticle.get());
   }
 
   @Test
```
kellyktan, line 134: You can just create the `when(webpageStream.read(...))` instead of creating a new one.

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
```diff
@@ -65,15 +65,15 @@
   private NewsContentExtractor realNewsContentExtractor = new NewsContentExtractor();
```
kellyktan, line 59: Reminder to remove any of these constants that are now obsolete.

kellyktan (changes requested):

info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java
KathyFeiyang: `realNewsContentExtractor` is used in this [test](https://github.com/googleinterns/voter-central/blob/5f23fd69571b9fb59ae42fb01f3d469ca032bf99/info-compiler/src/test/java/com/google/sps/NewsContentExtractorTest.java#L149) and the following one. Those two tests are meant to test if ContentExtractor can catch exceptions or null, so the parser is the real parser that reads a bad stream input

KathyFeiyang: The real parser doesn't read from any real webpages, it just calls on InputStream to read something, which will trigger an exception that is meant to be caught.

kellyktan: *I actually meant NEWS_ARTICLE_WITH_EMPTY_TITLE_CONTENT, etc. but couldn't comment up there since there wasn't a recent  change (sorry!)

KathyFeiyang: Oh gotcha! I missed those. Those are fixed in the commit just now.
Thank you!

info-compiler/src/test/java/com/google/sps/WebCrawlerTest.java
```diff
@@ -12,12 +12,215 @@
...
+    // consistent with those of {@code EXPECTED_NEWS_ARTICLE}.
+    // Since Mockito doesn't support the mocking of static methods, {@code NewsContentExtractor}'s 
+    // {@code extractContentFromHtml()} is not insular to this "unit" test. @TODO [Might modify
+    // {@code NewsContentExtractor} to aid test-driven development.
+    URL url = new URL(VALID_URL);
+    URL robotsUrl = new URL(url.getProtocol(), url.getHost(), "/robots.txt");
+    Grant grant = mock(Grant.class);
+    when(grant.hasAccess()).thenReturn(true);
+    when(grant.getCrawlDelay()).thenReturn(DELAY);
+    WebCrawler webCrawlerSpy = spy(webCrawler);
```
kellyktan: Are you using this anywhere?

info-compiler/src/test/java/com/google/sps/WebCrawlerTest.java
```diff
@@ -12,12 +12,215 @@
...
+    // EXPECTED_NEWS_ARTICLE}.
+    // Assume that the libraries {@code URL} and {@code RobotsTxt} work as intended.
+    // Since Mockito doesn't support the mocking of static methods, {@code NewsContentExtractor}'s 
+    // {@code extractContentFromHtml()} is not insular to this "unit" test. @TODO [Might modify
+    // {@code NewsContentExtractor} to aid test-driven development.
+    URL url = new URL(VALID_URL);
+    Optional<NewsArticle> potentialNewsArticle = webCrawler.scrapeAndExtractFromHtml(url);
+    Assert.assertTrue(potentialNewsArticle.isPresent());
+    Assert.assertEquals(EXPECTED_NEWS_ARTICLE, potentialNewsArticle.get());
+  }
```
kellyktan: nit: Don't test with real data.  Probably can just get rid of this test since pretty much all of the custom logic is in politelyscrapeAndExtractFromHtml anyways.

info-compiler/src/test/java/com/google/sps/WebCrawlerTest.java
```diff
@@ -12,12 +12,215 @@
...
+   * {@code webCrawler}. We choose to reset, instead of creating/destroying the Datastore emulator
+   * at each test is because {@code datastoreHelper.stop()} sometimes generates a {@code
+   * java.net.ConnectException}, when making HTTP requests. Hence we try to limit the number of
+   * times {@code datastoreHelper} is created/destroyed.
+   */
+  @Before
+  public void resetDatastore() throws IOException {
+    datastoreHelper.reset();
+    datastore = datastoreHelper.getOptions().getService();
+    webCrawler = new WebCrawler(datastore);
```
kellyktan: Maybe pass in a newsContentExtractor and/or relevancyChecker so you can mock them?

kellyktan (changes requested):

kellyktan (approved):
Looks great, thanks! Maybe for the remaining commented tests just add TODOs about what tests to add and stash the actual commented-out tests for now before merging this.

info-compiler/src/main/java/com/google/sps/webcrawler/NewsContentProcessor.java
```diff
@@ -19,11 +19,12 @@
...
-  private static final int MAX_WORD_COUNT = 100;
+  static final int MAX_WORD_COUNT = 100;
 
   // @TODO [Implement more advanced processing methods.]
   /** Extracts the first {@code MAX_WORD_COUNT} words from the news article content. */
-  public static NewsArticle process(NewsArticle newsArticle) {
+  public static NewsArticle process(NewsArticle originalNewsArticle) {
+    NewsArticle newsArticle = new NewsArticle(originalNewsArticle);
     String rawContent = newsArticle.getContent();
     String[] splitContent = rawContent.split(" ");
```
ganewman: Maybe you could do `newsArticle.getContent().split(" ")` since you don't use `rawContent` after this?

tjyoung99 (approved):

ganewman (approved):

PR #28, opened 2020-07-29
Extend RelevancyChecker to compute the salience of the party with which the candidate is affiliated. Modify associated interfaces and write unit tests.

KathyFeiyang:

**1. Extend `RelevancyChecker` to also compute the salience of the party with which the candidate is affiliated. Determine relevancy by checking if both the candidate name salience and the party name salience are bigger than their respective thresholds.**
- Will compute a meaningful `PARTY_NAME_THRESHOLD` using this [dataset](https://datasetsearch.research.google.com/search?query=%20news%20articles%20election&docid=YqB9uis7wW2pNi7aAAAAAA%3D%3D). Will compute an average party name salience among a test set of candidates and their news articles, and then evaluate this average value using a test set.

**2. Modify associated interfaces in `WebCrawler`, `InfoCompiler` and their tests.**

**3. Add unit tests for `RelevancyChecker`.**

### Is this fixing an issue?

none

none

- [ x ] Test written/updating
- [ x ] Tests passing


info-compiler/src/main/java/com/google/sps/data/State.java
```diff
@@ -0,0 +1,75 @@
...
+package com.google.sps.data;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** A class of state abbreviation and state name pairs. */
+public class State {
+  public static final Map<String, String> abbreviationToName = new HashMap<>(50);
+  static {
+    abbreviationToName.put("AL", "Alabama");
```
suparnabarua: Include a sanity test that you have 50 entries in this map :)

info-compiler/src/main/java/com/google/sps/infocompiler/InfoCompiler.java
```diff
@@ -320,11 +321,14 @@ void storeElectionContestInDatabase(String electionQueryId, JsonObject contest)
     List<Value<Boolean>> candidateIncumbency =
         new ArrayList<>(electionEntity.getList("candidateIncumbency"));
     // Obtain candidate information and create candidate entities in the database.
+    String state = electionEntity.getString("state");
+    String stateName = state.isEmpty() ? null : State.abbreviationToName.get(state);
```
suparnabarua: Is it possible for state to be invalid or lower case? In that case your map lookup would fail. If so, it would be good to convert state to uppercase before doing the lookup

suparnabarua (changes requested):

info-compiler/src/main/java/com/google/sps/data/State.java
KathyFeiyang: Will do!

info-compiler/src/main/java/com/google/sps/infocompiler/InfoCompiler.java
KathyFeiyang: The state is guaranteed to be two upper class letters or an empty string when it gets stored into the database (implementation can be viewed [here](https://github.com/googleinterns/voter-central/blob/kathy_extend_RelevancyChecker/info-compiler/src/main/java/com/google/sps/infocompiler/InfoCompiler.java#L284)). In InfoCompilerTest, `storeBaseElectionInDatabase_checkDatastoreEntityConstructionFromJsonWithState` and `storeBaseElectionInDatabase_checkDatastoreEntityConstructionFromJsonWithoutState` will check for that

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -24,13 +24,13 @@
...
 
-// @TODO [Compute the salience of other features, such as location, election, to improve accuracy.]
 /**
  * A utility class that performs entity analysis to check the relevancy of news article content to a
  * candidate.
  */
 public class RelevancyChecker {
-  static final double SALIENCE_THRESHOLD = 0.18;
+  static final double CANDIDATE_SALIENCE_THRESHOLD = 0.18;
+  static final double PARTY_SALIENCE_THRESHOLD = 0.0010;
```
kellyktan: nit: don't need trailing 0?

kellyktan (approved):
Is it possible to update the pull request title/description to reflect the state -> party name switch?

KathyFeiyang:
> Is it possible to update the pull request title/description to reflect the state -> party name switch?

Yes, done!

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -50,20 +50,26 @@ public RelevancyChecker() throws IOException {
...
+   * skips the salience checking for {@code partyName} and determines relevancy solely by looking
+   * at the salience of {@code candidateName}.
    */
-  public boolean isRelevant(NewsArticle newsArticle, String candidateName) {
-    double salience = computeSalienceOfName(newsArticle.getContent(), candidateName);
-    return salience >= SALIENCE_THRESHOLD;
+  public boolean isRelevant(NewsArticle newsArticle, String candidateName, String partyName) {
+    double candidateNameSalience = computeSalienceOfName(newsArticle.getContent(), candidateName);
+    double partyNameSalience =
+        (partyName == null)
```
kellyktan: is partyName == null the same as "NO PARTY AFFILIATION"?

KathyFeiyang: No, this is for when the party affiliation field doesn't exist in the query response. However, all responses I have seen provide the party affiliation field. I can remove this check!

kellyktan: Oh I was going to say to expand this check since it looks like there are several types of probably non-relevant parties which may skew the results: "NON PARTISAN", "NO PARTY AFFILIATION", "NONPARTISAN"
https://voter-central.ue.r.appspot.com/directory.html?address=default&listAllElections=true

KathyFeiyang: Oh that makes sense! Thank you I'll check for additional conditions.

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
```diff
@@ -50,20 +53,27 @@ public RelevancyChecker() throws IOException {
...
-    double salience = computeSalienceOfName(newsArticle.getContent(), candidateName);
-    return salience >= SALIENCE_THRESHOLD;
+  public boolean isRelevant(NewsArticle newsArticle, String candidateName, String partyName) {
+    double candidateNameSalience = computeSalienceOfName(newsArticle.getContent(), candidateName);
+    double partyNameSalience =
+        (partyName == null || partyName.equalsIgnoreCase(NO_PARTY_AFFILIATION)
+             || partyName.equalsIgnoreCase(NON_PARTISAN) || partyName.equalsIgnoreCase(NONPARTISAN))
+        ? PARTY_SALIENCE_THRESHOLD : computeSalienceOfName(newsArticle.getContent(), partyName);
+    return (candidateNameSalience >= CANDIDATE_SALIENCE_THRESHOLD
+        && partyNameSalience >= PARTY_SALIENCE_THRESHOLD);
```
ganewman: The way this is structured (returning a number that is exactly the threshold when you want to ignore party salience) is a little bit confusing. Maybe instead of using a ternary operator here you can just structure the whole block to use an if-else statement?

ganewman (changes requested):

info-compiler/src/main/java/com/google/sps/webcrawler/RelevancyChecker.java
KathyFeiyang: Ok I'll make the change! Thanks

tjyoung99 (approved):

ganewman (approved):

suparnabarua (approved):

PR #31, opened 2020-07-30
Implement data expiration and deletion mechanism, add lastModified field to Datastore entities, and write unit tests.

KathyFeiyang:

**1. Define data expiration as the time after a Datastore entity was created or last updated exceeding some threshold.**
- This threshold is set to be shorter than or equal to the cycle duration until `InfoCompiler` runs again.

**2. Query for and delete outdated Datastore entities, after the compilation process.**
- Election and candidate entities have data that come directly from the Civic Information API response. However, news article entities take noticeable time to compile. Thus, if we were to clear the entire database and then add new information, there would be a period of time when nothing can be hosted on the web application. _To make sure we can continuously serve information, we decided to first add new information, and then remove outdated information (and figure out a way to compute outdatedness)._
- For election and candidate entities, `InfoCompiler` overwrites existing elections and candidates if they exist, so duplication in adding & deleting is minimized.

**3. Set a new `lastModified` field for election and candidate entities in `InfoCompiler`, and for news article entities in `WebCrawler`.**

**4. Write instructions in README.**

### Is this fixing an issue?

none

none

- [ x ] Test written/updating
- [ x ] Tests passing

### DIFF
--- a/info-compiler/src/test/java/com/google/sps/InfoCompilerTest.java
+++ b/info-compiler/src/test/java/com/google/sps/InfoCompilerTest.java
@@ -64,6 +64,8 @@ import org.mockito.ArgumentCaptor;
 @RunWith(JUnit4.class)
 public final class InfoCompilerTest {
   private static final int ADDRESS_NUMBER = 957; // After screening.
+  private static String CORRECT_FORMAT_NAME = "Andrew Cuomo";
+  private static String CORRECT_FORMAT_NAME_IN_QUOTES = "\"Andrew Cuomo\"";
   private static final String ADDRESS = ",NY,New York,,,,,10028,,,,,East,,,84,Street,,,,144";
   private static final String STATE = "NY";
   private static final String NONTEST_ELECTION_QUERY_ID =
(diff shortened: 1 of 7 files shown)


info-compiler/src/test/java/com/google/sps/InfoCompilerTest.java
```diff
@@ -453,6 +475,69 @@ public void compileInfo_discardTestElection2000()
...
+    doCallRealMethod().when(infoCompilerMock).queryAndStoreElectionContestInfo();
+    doCallRealMethod()
+        .when(infoCompilerMock).queryAndStoreElectionContestInfo(anyString(), anyString());
+    doCallRealMethod().when(infoCompilerMock).queryAndStore(anyString(), anyString(), anyObject());
+    doCallRealMethod().when(infoCompilerMock).storeBaseElectionInDatabase(anyObject());
+    doCallRealMethod()
+        .when(infoCompilerMock).storeElectionContestInDatabase(anyString(), anyObject());
+    doCallRealMethod()
+        .when(infoCompilerMock)
+            .storeElectionContestCandidateInDatabase(anyObject(), anyObject(), anyObject());
```
kellyktan: Instead of  specifying every method to use the real method, you can use a mock spy and only override the behavior of `queryCivicInformation()`:
https://www.vogella.com/tutorials/Mockito/article.html#mockito_spy

kellyktan (changes requested):

info-compiler/src/test/java/com/google/sps/InfoCompilerTest.java
KathyFeiyang: Gotcha thank you! I'll use a mock spy.

info-compiler/src/test/java/com/google/sps/InfoCompilerTest.java
```diff
@@ -203,6 +205,9 @@ public void storeBaseElectionInDatabase_checkDatastoreEntityConstructionFromJson
     assertThat(electionEntity.getList("candidateIds")).isEmpty();
     assertThat(electionEntity.getList("candidateIncumbency")).isEmpty();
     assertThat(electionEntity.getString("state")).isEqualTo(STATE);
+    assertThat(((Timestamp) electionEntity.getValue("lastModified").get()).getSeconds()
```
suparnabarua: Why not compare the timestamp directly instead of doing .getSeconds()?

suparnabarua (changes requested):

info-compiler/src/test/java/com/google/sps/InfoCompilerTest.java
KathyFeiyang: It was initially because I didn't find a compareTo in its documentation

KathyFeiyang: This is fixed now!

kellyktan (approved):

ganewman (approved):

suparnabarua (approved):

tjyoung99 (approved):

